PR #50, opened 2017-06-19
Correcting multiple X-Frame-Options header

chekm8:
According to RFC7034, only these three values, DENY, SAMEORIGIN and ALLOW FROM are valid values and they are mutually exclusive; that is, the header field must be set to exactly one of these three values.   This will prevent the CSRF code from inserting it multiple times as well as duplicating it if it was already set elsewhere (e.g. IIS Header)

Addresses #7

### DIFF
--- a/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
+++ b/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
@@ -104,7 +104,7 @@ namespace System.Web.Helpers.AntiXsrf
                 // Adding X-Frame-Options header to prevent ClickJacking. See
                 // http://tools.ietf.org/html/draft-ietf-websec-x-frame-options-10
                 // for more information.
-                var xFrameHeaderName = "X-Frame-Options";
+                const string xFrameHeaderName = "X-Frame-Options";
                 if (httpContext.Response.Headers[xFrameHeaderName] == null)
                 {
                     httpContext.Response.AddHeader(xFrameHeaderName, "SAMEORIGIN");
--- a/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
+++ b/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
@@ -104,7 +104,11 @@ namespace System.Web.Helpers.AntiXsrf
                 // Adding X-Frame-Options header to prevent ClickJacking. See
                 // http://tools.ietf.org/html/draft-ietf-websec-x-frame-options-10
                 // for more information.
-                httpContext.Response.AddHeader("X-Frame-Options", "SAMEORIGIN");
+                var xFrameHeaderName = "X-Frame-Options";
+                if (httpContext.Response.Headers[xFrameHeaderName] == null)
+                {
+                    httpContext.Response.AddHeader(xFrameHeaderName, "SAMEORIGIN");
+                }
             }
 
             // <input type="hidden" name="__AntiForgeryToken" value="..." />
--- a/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
+++ b/src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
@@ -104,10 +104,10 @@ namespace System.Web.Helpers.AntiXsrf
                 // Adding X-Frame-Options header to prevent ClickJacking. See
                 // http://tools.ietf.org/html/draft-ietf-websec-x-frame-options-10
                 // for more information.
-                const string xFrameHeaderName = "X-Frame-Options";
-                if (httpContext.Response.Headers[xFrameHeaderName] == null)
+                const string frameHeaderName = "X-Frame-Options";
+                if (httpContext.Response.Headers[frameHeaderName] == null)
                 {
-                    httpContext.Response.AddHeader(xFrameHeaderName, "SAMEORIGIN");
+                    httpContext.Response.AddHeader(frameHeaderName, "SAMEORIGIN");
                 }
             }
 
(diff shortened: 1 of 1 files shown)


chekm8:
I was just checking back to see if there was anything additional I had to do for the pull to be reviewed/accepted.  This is my first contribution and I believe the license agreement was completed successfully.
Looking forward to this fix being incorporated, Thanks!

src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
```diff
@@ -104,7 +104,11 @@ public TagBuilder GetFormInputElement(HttpContextBase httpContext)
                 // Adding X-Frame-Options header to prevent ClickJacking. See
                 // http://tools.ietf.org/html/draft-ietf-websec-x-frame-options-10
                 // for more information.
-                httpContext.Response.AddHeader("X-Frame-Options", "SAMEORIGIN");
```
Eilon: Should this instead just be changed to:

```c#
httpContext.Response.Headers.Set("X-Frame-Options", "SAMEORIGIN");

```

This would make sure that anti-forgery gets the header it wants by overwriting any existing value.

Eilon:
BTW sorry for the long delay on this! I got sent some reminders from a couple of colleagues to get back to you on this.

src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
chekm8: Thanks for getting back to me Eilon.  The "If"/Add check was to ensure that say if an administrator in IIS forced this header to a setting other than SAMEORIGIN (e.g. DENY or ALLOW FROM) it would allow them to.  If you always just set it to SAMEORIGIN you are not giving the admin the option to override it.

chekm8: I was just checking back in, does the above comment make sense?

Eilon: Ok this makes sense. I guess it's debatable which one should win over the other but I don't feel strongly about that, so this should be fine.

Eilon:
@dougbu can you do an extra review and merge when ready?

src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
```diff
@@ -104,7 +104,11 @@ public TagBuilder GetFormInputElement(HttpContextBase httpContext)
...
+                var xFrameHeaderName = "X-Frame-Options";
```
Eilon: Oh actually one tiny comment: this should be a `const string` because it's not really a variable.

dougbu: Please also rename this `const` because the command-line build hits an FxCop error:
```
SA130: The variable name 'xFrameHeaderName' begins with a prefix that looks like Hungarian notation. Remove the prefix or add it to the list of allowed prefixes.
```

src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
```diff
@@ -104,7 +104,11 @@ public TagBuilder GetFormInputElement(HttpContextBase httpContext)
                 // Adding X-Frame-Options header to prevent ClickJacking. See
                 // http://tools.ietf.org/html/draft-ietf-websec-x-frame-options-10
                 // for more information.
-                httpContext.Response.AddHeader("X-Frame-Options", "SAMEORIGIN");
+                const string xFrameHeaderName = "X-Frame-Options";
```
dougbu: Name will still fail a build with the SA130 error. (Might be worth setting up AppVeyor for this repo.)

chekm8: @dougbu I uploaded the `string const` change to my repo, Do I need to initiate another pull request or anything?

Do you need further changes to get past the SA130 error?

dougbu: Yes, please renamed the `const` to something like `frameHeader` or `frameHeaderName`. That'll get command-line builds working.

chekm8: All set, let me know if you need anything else.  Thanks.

src/System.Web.WebPages/Helpers/AntiXsrf/AntiForgeryWorker.cs
```diff
@@ -104,7 +104,11 @@ public TagBuilder GetFormInputElement(HttpContextBase httpContext)
                 // Adding X-Frame-Options header to prevent ClickJacking. See
                 // http://tools.ietf.org/html/draft-ietf-websec-x-frame-options-10
                 // for more information.
-                httpContext.Response.AddHeader("X-Frame-Options", "SAMEORIGIN");
+                const string frameHeaderName = "X-Frame-Options";
```
dougbu: Shoot. Command-line build now fails with
```
SA1303: Constants must start with an upper-case letter: frameHeaderName.
```

Please rename this again -- `FrameHeaderName`. (I wasn't aware this error was enabled.)

chekm8: How about now :)

chekm8:
@dougbu Thanks,  I know I can grab a nightly build but, do you guy have a estimated timeline for your next RTM release and nuget package build? I'm not really sure how that process works.

dougbu:
@chekm8 the best place to see the future is https://github.com/aspnet/Home/wiki/Roadmap

The page needs a refresh now that 2.0.0 is out. But, I recommend checking there occasionally for more detail on 2.1.0.

PR #51, opened 2017-06-22
Revert "Remove IDisposable usage of `TestParserManager`."

dougbu:
- reverts commit 55f1ff5
- 55f1ff5 was likely an accidental revert of part of 2e98bc0
  - in turn, 2e98bc0 avoided a leak (and sometimes an eventual hang) when testing

Add `global.json` to avoid problems building on systems with newer tooling


dougbu:
FYI we don't seem to have AppVeyor set up here. Builds fine on my CI-like machine.

PR #71, opened 2017-10-17
Add NetStandard test assembly to SkipStrongNames.xml

dougbu:


dougbu:
:ok: this does fix AppVeyor builds 🎉

PR #73, opened 2017-10-19
Quick fix: Get .NET Standard test project testing in Release configuration

dougbu:
- solution mapped Release configuration to a Debug build of the .NET Standard test project


dougbu:
d2aee97b

Issue #72, opened 2017-10-17
16 test failures related to `JObject` instances cast to `dynamic`


dougbu:
Test failures do not occur on all machines. But, they happen consistently on machines with .NET 4.7 installed. All failures involve unexpected `RuntimeBinderException`s. And, those `Exception`s are thrown in consistent locations:
```
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException : 'Newtonsoft.Json.Linq.JObject' does not contain a definition for 'oauth_token'
    at CallSite.Target(Closure , CallSite , Object )
    ...
```
The error messages mention `oauth_token` in two cases (in `Microsoft.AspNet.Facebook.Test.FacebookContextModelBinderTest`) and `ExceptionType` in the remaining 14 failures (2 in `System.Web.Http.ExceptionHandlingTest` and 12 in `System.Web.Http.IncludeErrorDetailTest`).

Could workaround the `ExceptionType` failures, using something other than `dynamic` in the test code. However the `FacebookClient`'s `ParseSignedRequest(...)` method internally casts a `JObject` instance to `dynamic`. (Unclear why e.g. `FacebookAuthorizeFilterHookTest` tests succeed while `FacebookContextModelBinderTest` tests fail.)

dougbu:
Self-assigning since this gets in my way

dougbu:
e4c8e01a

Issue #65, opened 2017-10-09
Update to latest tooling


dougbu:
Bring this repo up-to-date w.r.t. the MSBuild version, the StyleCop version, and the Castle.Core / Moq / xUnit versions used in tests. Do not change product code, dependencies, or runtime targets.

This includes a small quirk on the test side: Latest xUnit packages depend on various .NET Standard packages if the test project targets anything earlier than .NET 4.5.2. Update the test projects to target .NET 4.5.2 to avoid those dependencies. We're only testing on more-recent Windows, and therefore desktop .NET, versions anyhow.

See PR aspnet/WebHooks#167 for a smaller variant of this work.

dougbu:
Self-assigning because using the old tooling is slowing work on my local machine and may be the root cause of odd test failures that don't reproduce on the (older) CI infrastructure.

dougbu:
StyleCop update: 2c759d34

dougbu:
Newtonsoft.Json upgrade: e4c8e01a

dougbu:
MSBuild and xUnit MSBuild upgrade: 634c9d28
Testing .NET Framework upgrade (enabling Moq / xUnit upgrades): 4b309584

dougbu:
Moq and Castle.Core upgrade: 62cc8b3f

dougbu:
xUnit runtime upgrade: 64806c4

dougbu:
Add xUnit analyzers and take fixer suggestions:
- 74fcc242
- 76743944
- f08460bc
- 66e893de
- dca17706d8

xUnit MSBuild runner upgrade to stable version: 2d183608
xUnit VisualStudio runner upgrade: 125b71ac

PR #74, opened 2017-10-19
Move tests to Newtonsoft.Json v7.0.1

dougbu:
- #72 and also related to #65
- v7.0.1 is the next major version after what the product code is built against (v6.0.4)
  - this version corrects test failures I saw on a machine with .NET 4.7 (#72)
- add binding redirects and include the app.config files in test projects
- avoid new `TypeLoadException`s for Newtonsoft.Json that break numerous tests
  - consistently disable Copy Local (aka `%(Private)` metadata) in `src` projects
  - add Newtonsoft.Json references in test projects that bring it in transitively
    - otherwise, product code is last to mention Newtonsoft.Json assembly


dougbu:
e4c8e01a

PR #78, opened 2017-10-20
Minor improvements

dougbu:
- ignore launchSettings.json files
- fix test project's `%(HintPath)' metadata for Facebook assembly
- disable shadow copy in all xUnit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ _[Ss]cripts
 *.sln.ide
 .vs/
 project.lock.json
+*launchSettings.json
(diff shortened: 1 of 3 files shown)


Eilon:
BTW, not that I have a problem with it, but why disable shadow copying?

dougbu:
> why disable shadow copying?

App domains are enabled when using the xUnit MSBuild runner, making the setting relevant when testing outside VisualStudio. Unfortunately, shadow copying doesn't copy enough files in a few odd cases. This was a longstanding problem but it doesn't affect most builds. (Can't remember the exact "odd cases" at the moment.)

In any case, sticking with the single test folder doesn't hurt.

dougbu:
66625656

PR #80, opened 2017-10-20
Move to MSBuild v15.0 and latest xUnit MSBuild runner

dougbu:
- move to MSBuild v15.0
- use only VS 2017 or later
- move to latest version of the xUnit MSBuild runner

See descriptions of individual commits for details.

### DIFF
--- a/Runtime.sln
+++ b/Runtime.sln
@@ -1,8 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30501.0
-MinimumVisualStudioVersion = 10.0.40219.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.27016.1
+MinimumVisualStudioVersion = 15.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{A9836F9E-6DB3-4D9F-ADCA-CF42D8C8BA93}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{C40883CD-366D-4534-8B58-3EA0D13136DF}"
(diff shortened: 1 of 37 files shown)


Runtime.sln
```diff
@@ -1,8 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30501.0
-MinimumVisualStudioVersion = 10.0.40219.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.27016.1
+MinimumVisualStudioVersion = 15.0
```
dougbu: FYI I edited the .sln files then made temporary updates in VS to see if it wanted anything else e.g. the `$(SolutionGuid)` additions.

build.cmd
```diff
@@ -22,20 +43,20 @@ if %ERRORLEVEL% neq 0 goto BuildFail
 goto BuildSuccess
 
 :BuildDefaults
-%MSBuild% Runtime.msbuild /m /nr:false /p:Platform="Any CPU" /p:Desktop=true /v:M /fl /flp:LogFile=bin\msbuild.log;Verbosity=detailed
+%MSBuild% Runtime.msbuild /m /nr:false /p:Platform="Any CPU" /p:Desktop=true /v:M /fl /flp:LogFile=bin\msbuild.log;Verbosity=Normal
```
dougbu: This verbosity is now consistent with the command a few lines above (the non-default build).

tools/WebStack.tasks.targets
```diff
@@ -178,7 +187,7 @@
 
                             if (failures > 0)
                             {
-                                foreach (XElement classWithFailure in assembly.Elements(XName.Get("class"))
+                                foreach (XElement classWithFailure in assembly.Elements(XName.Get("collection"))
```
dougbu: Tied to move from `XmlV1` to `Xml` results file. The few bits of the files we care about don't change much.

tools/WebStack.xunit.targets
```diff
@@ -3,15 +3,13 @@
 
     <!-- This is a separate MSBuild file so that we can survive upgrades of the xunit NuGet package
          and also still work with NuGet Package Restore. -->
-    <ItemGroup>
-        <XunitMsBuildRunner Include="..\packages\**\xunit.runner.msbuild.dll"/>
-    </ItemGroup>
-    <UsingTask TaskName="Xunit.Runner.MSBuild.xunit" AssemblyFile="@(XunitMsBuildRunner)"/>
+    <Import Project="..\packages\**\xunit.runner.msbuild.props"/>
```
dougbu: xUnit made a breaking change in the name of their main assembly for this runner. Using the .props file avoids that change and should hopefully avoid breaking changes in the future (unless xUnit starts shipping multiple versions of this file 😈).

dougbu:
#83 replaces this.

PR #81, opened 2017-10-20
Build tests targeting .NET 4.5.2

dougbu:
- precondition for other tool upgrades; xUnit in particular does not support older framework versions

### DIFF
--- a/tools/WebStack.settings.targets
+++ b/tools/WebStack.settings.targets
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import
+      Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props"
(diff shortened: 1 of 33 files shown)


tools/WebStack.settings.targets
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import
+      Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props"
+      Condition="'$(MicrosoftCommonPropsHasBeenImported)' != 'true' AND Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')"/>
```
dougbu: This standard `<import/>` brings in the Directory.Build.props file, if any. If left out, that file doesn't get brought in until the end of our project files i.e. when Microsoft.CSharp.targets (or similar) is imported. Such a delay is confusing and makes Directory.Build.props confusing to use.

dougbu:
#84 replaces this.

PR #85, opened 2017-10-23
Upgrade to Moq v4.7.142 and Castle.Core v4.2.1

dougbu:
- part of #65
- NuGet added System.Configuration references

- add `[InternalsVisibleTo("DynamicProxyGenAssembly2, ...)]` attribute in MVC to fix more than 70 failures
  - mainly impacted mocks of `internal` interface implementations e.g. `ActionDescriptor` & `IUniquelyIdentifiable`
- split some `Setup(...)` and `SetupGet(...)` lambda expressions in two
  - mainly impacted mocks of types with `protected` constructors e.g. `HttpResponseBase`
- typo in `HttpRouteCollectionTest` should have always broken the test
- not sure why tests using `PortReserver` weren't failing before; tests shouldn't be opening UDP ports

### DIFF
--- a/src/System.Web.Mvc/Properties/AssemblyInfo.cs
+++ b/src/System.Web.Mvc/Properties/AssemblyInfo.cs
@@ -4,7 +4,6 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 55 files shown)


dougbu:
62cc8b3f

PR #88, opened 2017-10-25
Add xUnit analyzers and take fixer suggestions (1 of 5): Common to System.Web.Cors.Test

dougbu:
- part of #65
- do not use `xunit.core` package in Microsoft.TestCommon
- improve `TestData` behavior when tests sharing dataset properties run in parallel

Manual changes:
- work around xUnit1013, Public method should be marked as test
  - move duplicated `[Theory]` and `[TestDataSet(...)]` attributes down to `abstract` methods
  - add suppression due to xnuit/xunit#1466
- work around xUnit1026, `[Theory]` method doesn't use all parameters
  - add new data sets or `GC.KeepAlive(...)`
- reset `OAuthWebSecurity.OAuthDataProvider` after use in two tests

### DIFF
--- a/test/Common/AttributeListTest.cs
+++ b/test/Common/AttributeListTest.cs
@@ -54,14 +54,14 @@ namespace System.ComponentModel
         public void AttributeListContainsWrappedTrue()
         {
             Attribute presentAttribute = _collection[2];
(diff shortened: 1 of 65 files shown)


dougbu:
74fcc242

PR #89, opened 2017-10-25
Add xUnit analyzers and take fixer suggestions (2 of 5): System.Web.Helpers.Test to System.Web.Http.SignalR.Test

dougbu:
- part of #65

Few manual changes:
- suppress xUnit1013, Public method should be marked as test
  - mark a few methods as `private` for the same reason
- work around xUnit1026, `[Theory]` method doesn't use all parameters
  - add `GC.KeepAlive(...)`


dougbu:
76743944

PR #91, opened 2017-10-25
Add xUnit analyzers and take fixer suggestions (4 of 5): System.Web.Http.Tracing.Test to System.Web.Mvc.Test

dougbu:
- part of #65

Few manual changes:
- fix xUnit1026, "`[Theory]` method doesn't use all parameters" issues
  - turn `[Theory]` that used no parameters into a `[Fact]`
  - add new dataset properties; see `EncodedDataSets` in particular
- `Assert.Equal<T>(...)` -> `Assert.Equal(...)`
- avoid Linq's `.Where` inside `Assert.Contains(...)` and similar


dougbu:
66e893de

PR #92, opened 2017-10-25
Add xUnit analyzers and take xUnit fixer suggestions (5 of 5): System.Web.Razor.Test to end

dougbu:
- part of #65

Few manual changes:
- work around xunit/xunit#1502
- suppress xUnit1013, Public method should be marked as test
  - helper method is called from multiple test classes; none inherit from defining class
- fix xUnit1026, `[Theory]` method doesn't use all parameters
  - add new data set properties
- suppress xUnit2013, Do not use equality check to check for collection size.
  - calls `Count` because the `Mock<T>` does not set up `GetEnumerable()`
- `Assert.Equal<T>(...)` -> `Assert.Equal(...)`
- avoid Linq's `.Where` inside `Assert.Contains(...)` and similar

### DIFF
--- a/test/System.Web.Razor.Test/Text/LookaheadTextReaderTestBase.cs
+++ b/test/System.Web.Razor.Test/Text/LookaheadTextReaderTestBase.cs
@@ -159,9 +159,7 @@ namespace System.Web.Razor.Test.Text
             Assert.Equal(expectedRaw, actualLocation.AbsoluteIndex);
             Assert.Equal(expectedChar, actualLocation.CharacterIndex);
             Assert.Equal(expectedLine, actualLocation.LineIndex);
-
-            // Workaround xunit/xunit#1502
-            Assert.Equal('b' - '\0', reader.Peek());
+            Assert.Equal('b', reader.Peek());
             Assert.Equal(read, readMethod(reader));
         }
 
@@ -190,9 +188,7 @@ namespace System.Web.Razor.Test.Text
             Assert.Equal(0, actualLocation.LineIndex);
             Assert.Equal(1, reader.CurrentLocation.CharacterIndex);
             Assert.Equal(0, reader.CurrentLocation.LineIndex);
-
-            // Workaround xunit/xunit#1502
-            Assert.Equal('b' - '\0', reader.Peek());
+            Assert.Equal('b', reader.Peek());
         }
 
         private static void RunAll(Action<StringBuilder, LookaheadTextReader>[] readerCommands, StringBuilder builder, LookaheadTextReader reader)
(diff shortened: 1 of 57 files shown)


test/System.Web.Razor.Test/Text/LookaheadTextReaderTestBase.cs
```diff
@@ -159,7 +159,9 @@ protected void RunReadUntilTest(Func<LookaheadTextReader, string> readMethod, in
             Assert.Equal(expectedRaw, actualLocation.AbsoluteIndex);
             Assert.Equal(expectedChar, actualLocation.CharacterIndex);
             Assert.Equal(expectedLine, actualLocation.LineIndex);
-            Assert.Equal('b', reader.Peek());
+
+            // Workaround xunit/xunit#1502
+            Assert.Equal('b' - '\0', reader.Peek());
```
Eilon: This is really gross. Should we just disable the affected analyzer in the meantime? We should not change our code to work around bugs in unnecessary components...

test/System.Web.WebPages.Test/Html/HtmlHelperTest.cs
```diff
@@ -195,13 +195,15 @@ private static object GetAttributes()
...
         }
 
+#pragma warning disable xUnit1013 // Public method should be marked as test
         /// <summary>
         /// Will invoke a helper with overload that accepts custom attribute with a name containing
         /// and underscore as an anonymous object, and will then assert that the resulted html
         /// has the attribute name underscore correctly transformed to a dash
         /// </summary>
         /// <param name="helperInvocation"></param>
         public static void AssertHelperTransformsAttributesUnderscoresToDashs(Func<HtmlHelper, object, IHtmlString> helperInvocation)
```
Eilon: Can this just be `internal`?

test/System.Web.Razor.Test/Text/LookaheadTextReaderTestBase.cs
dougbu: Tried suppressing the analyzer that's blowing up due to xunit/xunit#1502 but unfortunately it still runs far enough to hit the NRE before noticing the rule is disabled. Could either leave this alone or disable the xUnit analyzers entirely in this project. This would be temporary in any case, lasting only until we address #96.

Which option do you prefer?

Eilon: I'd rather not have the analyzer because we don't super need that, and we really don't want this ugly code.

dougbu:
:up::date: to address comments and to fix the CA2227 violation introduced in eebbf3884

dougbu:
dca17706d8 and 15169a59e9

PR #94, opened 2017-10-26
Upgrade xUnit Visual Studio runner to latest stable version (v2.3.0)

dougbu:
- enable app domains in all testing; `denied` now breaks testing in VS
  - VS runner previously ignored this setting; see xunit/xunit#1533
- remove `"shadowCopy":  false` setting too
  - with updated xUnit packages, this setting slows testing down significantly


dougbu:
:up::date: to v2.3.1.

dougbu:
Oops, this one was 125b71ac

PR #109, opened 2017-11-28
Avoid odd test failures

dougbu:
- XML serializers handle `DBNull` on some machines but not others
- possible an underlying problem was fixed recently, in .NET Core SDK 2.0.3
- but, better to make builds work almost everywhere and wait for .NET Core SDK 2.1.x (where `DBNull` is seralizable)


dougbu:
@Eilon this time it's four changed lines 😃

Can't really test end-to-end 'til this goes into `master`. AppVeyor worked last time around, as did local builds on multiple machines ☹️ Good news is our most recent official CI build was otherwise clean.

dougbu:
86f6078419

Issue #120, opened 2018-01-23
ValidationContext.MemberName is set incorrectly in the DataAnnotationsModelValidator


mkArtakMSFT:
Problem Description
=================================
My customer creating custom validation class which deliveerd from System.ComponentModel.DataAnnotations.ValidationAttribute.
Using this in .NET4.7.1 ASP.NET Web API 2.0, ValidationContext.MemberName is set "DisplayName".
This is problem for custom class because they can not know MemberName without this property.

- Expected behavior
    ValidationContext.MemberName is "TestValue3"
- Actual behavior
    ValidationContext.MemberName is "Display name From 'DisplayAttribute'"

TestValue3 is set "DisplayAttribute".

```
    public class TestModel
    {
        [Validations.CustomValidation]
        public string TestValue1 { get; set; }
        [Validations.CustomValidation(ErrorMessage = "ErrorMessage: {0} ErrorMessage")]
        public string TestValue2 { get; set; }
        [Validations.CustomValidation(ErrorMessage = "ErrorMessage: {0} ErrorMessage")]
        [Display(Name = "Display name From 'DisplayAttribute'")]
        public string TestValue3 { get; set; }
        public string Message { get; set; }
    }
```

mkArtakMSFT:
@dougbu, please associate the fix we've discussed with this bug. //cc: @danroth27

dougbu:
For historical interest: Was a PR for this in the old CodePlex repo http://aspnetwebstack.codeplex.com/SourceControl/network/forks/astolfoho/fixWebApi/contribution/8918. But, I can't find an earlier issue on this.

dougbu:
Moved to 3.2.5 (really 3.2.5-preview1).

dougbu:
9829dd8225

dougbu:
This fix is in MVC 5.2.5 preview 1 and later, not 5.2.4.

FYI the commit date may be much earlier than when the changes are merged into the `master` branch.

PR #121, opened 2018-01-23
Give `ValidationContext.MemberName` a distinct value from `DisplayName`

dougbu:
- #120

No longer true:
~~`ValidationAttribute` classes that (incorrectly) expected `MemberName` to always be set or that
instantiate `ValidationResult`s using the `DisplayName` may be negatively impacted. For exmaple,
`MemberName` will be `null` when binding parameters and items in a collection.~~

### DIFF
--- a/src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
+++ b/src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
@@ -33,11 +33,11 @@ namespace System.Web.Http.Validation.Validators
             // Per the WCF RIA Services team, instance can never be null (if you have
             // no parent, you pass yourself for the "instance" parameter).
             
-            string memberName = metadata.GetDisplayName();
+            var memberName = metadata.PropertyName;
             ValidationContext context = new ValidationContext(container ?? metadata.Model)
             {
-                DisplayName = memberName,
-                MemberName = memberName
+                DisplayName = metadata.GetDisplayName(),
+                MemberName = memberName,
             };
 
             ValidationResult result = Attribute.GetValidationResult(metadata.Model, context);
--- a/src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
+++ b/src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.ComponentModel.DataAnnotations;
+using System.Configuration;
 using System.Linq;
 using System.Web.Http.Metadata;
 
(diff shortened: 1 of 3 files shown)


src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
```diff
@@ -33,11 +33,11 @@ public override IEnumerable<ModelValidationResult> Validate(ModelMetadata metada
             // Per the WCF RIA Services team, instance can never be null (if you have
             // no parent, you pass yourself for the "instance" parameter).
             
-            string memberName = metadata.GetDisplayName();
+            var memberName = metadata.PropertyName;
```
dougbu: For a while, aspnet/MVC's version of this file used `metadata.PropertyName ?? metadata.ModelType.Name`. But, though that's more like what `GetDisplayName()` did, it's also incorrect. We fixed it to look like the above in aspnet/Mvc@aecfe778ee.

dougbu: I'm on the fence about this particular expression. In favour of a change: We recognized a problem and fixed that problem in ASP.NET Core MVC. Against: MVC 5.2.4 still uses `metadata.PropertyName ?? metadata.ModelType.Name` i.e. the consistency across ASP.NET argument. And, going this way makes it a larger change i.e. `MemberName` will now ignore both `[DisplayName("some name")]` and the type name for non-properties.

dougbu:
:up::date: with more tests and less trailing whitespace

src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
```diff
@@ -32,23 +32,23 @@ public override IEnumerable<ModelValidationResult> Validate(ModelMetadata metada
         {
             // Per the WCF RIA Services team, instance can never be null (if you have
             // no parent, you pass yourself for the "instance" parameter).
-            
-            string memberName = metadata.GetDisplayName();
+
+            var memberName = metadata.PropertyName;
```
dougbu: For a while, aspnet/MVC's version of this file used `metadata.PropertyName ?? metadata.ModelType.Name`. But, though that's more like what `GetDisplayName()` did, it's also incorrect. We fixed it to look like the above in aspnet/Mvc@aecfe778ee.

I'm on the fence about this particular expression. In favour of a change: We recognized a problem and fixed that problem in ASP.NET Core MVC. Against: MVC 5.2.4 still uses `metadata.PropertyName ?? metadata.ModelType.Name` i.e. the consistency across ASP.NET argument. And, going this way makes it a larger change i.e. `MemberName` will now ignore both `[DisplayName("some name")]` and the type name for non-properties.

pranavkm: Out of curiosity, when is `metadata.PropertyName` null? When we're enumerating over collection types?

dougbu: That and when we're validating a parameter. IOW whenever we're not dealing w/ a property.

src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
```diff
@@ -21,6 +29,13 @@ public DataAnnotationsModelValidator(IEnumerable<ModelValidatorProvider> validat
             Attribute = attribute;
         }
 
+        // Internal for testing
+        internal static bool UseLegacyValidationMemberName
```
dougbu: `static` only because an instance field for testing seemed wasteful.

I doubt it's necessary but could make this property `public` for Microsoft.AspNet.WebApi.Owin users, whose applications are less likely to involve `System.Configuration`. Could even move this (still `static`) into `HttpConfiguration` for high visibility. Thoughts?

dougbu:
:up::date:
/cc @Eilon since you wanted to have a gander.

src/System.Web.Http/Validation/Validators/DataAnnotationsModelValidator.cs
Eilon: Naah it's fine the way it is.

dougbu:
9829dd8225

Issue #135, opened 2018-03-03
Do not use `FormattedModelValue` in password editor template


dougbu:
- see aspnet/Mvc#7418
- make `@Html.Editor(expression, templateName: "password")` behave more like `@Html.Password(expression)`

dougbu:
2b7ac74060

PR #136, opened 2018-03-03
Do not use `FormattedModelValue` in password editor template

dougbu:
- #135
- see also PR aspnet/Mvc#7430
- add quirk switch to reverse this if necessary

### DIFF
--- a/test/System.Web.Mvc.Test/Html/Test/DefaultEditorTemplatesTest.cs
+++ b/test/System.Web.Mvc.Test/Html/Test/DefaultEditorTemplatesTest.cs
@@ -731,12 +731,61 @@ namespace System.Web.Mvc.Html.Test
         public void PasswordTemplateTests()
         {
             Assert.Equal(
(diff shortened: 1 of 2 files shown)


test/System.Web.Mvc.Test/Html/Test/DefaultEditorTemplatesTest.cs
```diff
@@ -731,14 +731,63 @@ public void ObjectTemplateDisplaysNullDisplayTextWithNullModelAndTemplateDepthGr
         public void PasswordTemplateTests()
         {
             Assert.Equal(
-                "<input class=\"text-box single-line password\" id=\"FieldPrefix\" name=\"FieldPrefix\" type=\"password\" value=\"Value\" />",
+                "<input class=\"text-box single-line password\" id=\"FieldPrefix\" name=\"FieldPrefix\" type=\"password\" />",
```
dougbu: This test (well, two tests in one) and the two theories using the data set below set `Model` and `FormattedModelValue`. I left similar settings in `PasswordTemplate_ReturnsInputElement_IgnoresValues()` due to its narrower intent.

dougbu:
2b7ac74060

Issue #117, opened 2018-01-08
ExpressionHelper contains a compatibility bug with Roslyn compiler


doggy8088:
There is a bug in the ASP.NET MVC 5.2.3 for a long time.   It's hard to explain this in English.  So I made a few steps to reproduce this problem.

# STEPS TO REPRODUCE THE PROBLEM

1. Create a simple ASP.NET MVC 5 project using Visual Studio 2017
2. Apply change to the follow code snippets.  Only 3 files to change.
   https://gist.github.com/doggy8088/58dd3c773ef9877155b19ebb2fda8525
3. Start debugging and navigate to /Home/Index page.  Then you will see the result like below:
   ![image](https://user-images.githubusercontent.com/88981/34677022-c5b3ea6c-f4c9-11e7-8d81-fc553696ff4b.png)
As you can see, the `@Html.TextBoxFor(m => data[i].Name)` generated field name in the for-loop in the view are all wrong.  

# MORE INFO

This problem is only exists when ASP.NET MVC 5 project using Roslyn to compiler views.

I was wrote an article about 2 years ago.  My article is written in Chinse.  If some of you can read Chinese, you can also reference my article over here: https://blog.miniasp.com/post/2016/03/06/ASPNET-MVC-5-View-Roslyn-problem-workaround.aspx

I actually proposed a [pull request](https://aspnetwebstack.codeplex.com/SourceControl/network/forks/doggy8088/ExpressionHelperFixed/changeset/1f069a92741a58f6410fe1b0773154f5c37c9257) to fix this problem about 2 years ago too.  You may check it out too.

mkArtakMSFT:
@dougbu, can you please look into this?

IanKemp:
tl;dr backporting the fixes from [this commit](MVChttps://github.com/aspnet/Mvc/commit/a045324d3a7bd29f030a120d2edd90fedb3fb5be) in ASP.NET Core should resolve the issue.

doggy8088:
@mkArtakMSFT Any new progress on this issue?

mkArtakMSFT:
@dougbu, did you get a chance to look into this?

dougbu:
@mkArtakMSFT I read the comments over the weekend and no longer think we need an investigation.

We fixed a related problem in ASP.NET Core and can do something similar here. That should fix the problem. Should however spend some time adding a number of tests covering more-complicated lambda expressions in this repo.

[...]

PR #140, opened 2018-03-12
Do not include compiler-generated names in expression names

dougbu:
- #117
- see also aspnet/Mvc@a045324d3a and PR aspnet/Mvc#3027, the fix for aspnet/Mvc#2890

nits: copy some improved comments from aspnet/Mvc@a045324d3a

### DIFF
--- a/src/System.Web.Mvc/ExpressionHelper.cs
+++ b/src/System.Web.Mvc/ExpressionHelper.cs
@@ -35,6 +35,7 @@ namespace System.Web.Mvc
 
                     if (!IsSingleArgumentIndexer(methodExpression))
                     {
+                        // Unsupported
                         break;
                     }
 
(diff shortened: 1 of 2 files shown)


src/System.Web.Mvc/ExpressionHelper.cs
```diff
@@ -59,7 +60,18 @@ public static string GetExpressionText(LambdaExpression expression)
                 else if (part.NodeType == ExpressionType.MemberAccess)
                 {
                     MemberExpression memberExpressionPart = (MemberExpression)part;
-                    nameParts.Push("." + memberExpressionPart.Member.Name);
+                    var name = memberExpressionPart.Member.Name;
+
+                    // If identifier contains "__", it is "reserved for use by the implementation" and likely compiler-
+                    // or Razor-generated e.g. the name of a field in a delegate's generated class.
+                    if (name.Contains("__"))
```
pranavkm: Would we need a compat switch to allow this? Seems like a super corner case though.

pranavkm: Presumably Razor doesn't stick a `CompilerGeneratedAttribute` or something along those lines on the member that we could use to infer this?

dougbu: We're already dealing with an infrequent case. Imaging a scenario where this fix _causes_ a problem is pretty extreme. I don't think the slight change is worth a compat switch. But, I'll ask around&hellip;

Roslyn does apply `[CompilerGenerated]` to its generated classes but not to fields in those classes. So, that attribute doesn't help distinguish `<>c__DisplayClass3_0.collection` (which is fine) from `<>c__DisplayClass5_1,CS$<>8__locals1` (which isn't). and `__` is what docs say. (See the last paragraph in [this section](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#identifiers) of the language spec for example.)

pranavkm:
:shipit:

dougbu:
b32ab42712

Issue #173, opened 2018-07-02
DefaultHttpBatchHandler creates innerRequest with scheme http even when outer batch request scheme is https


Rich-biomni:

dougbu:
dcee8e6a23

With many thanks to @JamesSinclairBiomni

PR #176, opened 2018-07-05
fix for https://github.com/aspnet/AspNetWebStack/issues/173

JamesSinclairBiomni:
This change will use the uri scheme of the outer request for all inner requests.

### DIFF
--- a/src/System.Web.Http/Batch/DefaultHttpBatchHandler.cs
+++ b/src/System.Web.Http/Batch/DefaultHttpBatchHandler.cs
@@ -187,7 +187,7 @@ namespace System.Web.Http.Batch
             foreach (HttpContent httpContent in streamProvider.Contents)
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                HttpRequestMessage innerRequest = await httpContent.ReadAsHttpRequestMessageAsync(request.RequestUri.Scheme);
+                HttpRequestMessage innerRequest = request.RequestUri == null ? await httpContent.ReadAsHttpRequestMessageAsync() : await httpContent.ReadAsHttpRequestMessageAsync(request.RequestUri.Scheme);
                 innerRequest.CopyBatchRequestProperties(request);
                 requests.Add(innerRequest);
             }
--- a/src/System.Web.Http/Batch/DefaultHttpBatchHandler.cs
+++ b/src/System.Web.Http/Batch/DefaultHttpBatchHandler.cs
@@ -187,7 +187,7 @@ namespace System.Web.Http.Batch
             foreach (HttpContent httpContent in streamProvider.Contents)
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                HttpRequestMessage innerRequest = await httpContent.ReadAsHttpRequestMessageAsync();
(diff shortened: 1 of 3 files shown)


src/System.Web.Http/Batch/DefaultHttpBatchHandler.cs
```diff
@@ -187,7 +187,7 @@ public virtual Task<HttpResponseMessage> CreateResponseMessageAsync(IList<HttpRe
             foreach (HttpContent httpContent in streamProvider.Contents)
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                HttpRequestMessage innerRequest = await httpContent.ReadAsHttpRequestMessageAsync();
+                HttpRequestMessage innerRequest = await httpContent.ReadAsHttpRequestMessageAsync(request.RequestUri.Scheme);
```
dougbu: To avoid changes to unrelated tests, please make this `null`-conditional e.g. `request.RequestUri?.Scheme`. If the repo disallows new-ish C# features, use a ternary expression.

test/System.Web.Http.Test/Batch/DefaultHttpBatchHandlerTest.cs
```diff
@@ -281,6 +283,29 @@ public Task ParseBatchRequestsAsync_Throws_IfRequestIsNull()
...
+        [Fact]
+        public async Task ParseBatchRequestsAsync_Returns_RequestsFromMultipartContent_WithUriSchemeSet_FromRequest()
+        {
+            DefaultHttpBatchHandler batchHandler = new DefaultHttpBatchHandler(new HttpServer());
+            HttpRequestMessage request = new HttpRequestMessage
+            {
+                Content = new MultipartContent("mixed")
+                {
+                    new HttpMessageContent(new HttpRequestMessage(HttpMethod.Get, "http://example.com/")),
+                    new HttpMessageContent(new HttpRequestMessage(HttpMethod.Post, "https://example.com/values"))
```
dougbu: Please add a part using a path-only relative URI e.g. `/api/values` since those are recommended and seem to be more common.

JamesSinclairBiomni: I can't supply a relative Uri within a batch request. If I do so then ParseBatchRequestsAsync calls ReadAsMultipartAsync (an extension on HttpContent) and that will throw the following error.

Message: System.IO.IOException : Error reading MIME multipart body part.
---- System.InvalidOperationException : This operation is not supported for a relative URI.

Does this suggest that its absolute uri's by design or is this a different issue?

src/System.Web.Http/Batch/DefaultHttpBatchHandler.cs
JamesSinclairBiomni: Done using a ternary. It's C# 5 so no ?.

test/System.Web.Http.Test/Batch/DefaultHttpBatchHandlerTest.cs
dougbu: Odd. But, if that's what enforced, so be it. On the off chance this is a bug I should open separately, please paste the full stack trace into this PR.

Issue #147, opened 2018-03-28
Make exception handling policy configurable for CORS


itrofimow:
In project i am working at custom exception handling OwinMiddleware is used as first in the pipeline, which catches every exception thrown by application and maps it to httpResponse/statusCode based on some logic. We replace IExceptionHanlder with a handler that just rethrows , thus we have a single global exception hanlder(middleware). The problem arised when we tried to add CORS - its System.Web.Http.Cors.CorsMessageHandler.SendAsync swallows exeptions and processes them by its own logic and that just broke plenty of things. 

We ended up with basically copy/paste of that Handler with SendAsync not catching everything, then we had to replace CorsMessageHanlder at HttpConfigurations.MessageHanlders after EnsureInitialized has been called and that looks pretty nasty.

So what i am asking for - is it possible to make that exception handling policy configurable?
Or did i miss something and it already is?

mkArtakMSFT:
Thanks for contacting us, @itrofimow,
@blowdart, @Tratcher any suggestions?

Tratcher:
Yeah, this try/catch is capturing way too much:
https://github.com/aspnet/AspNetWebStack/blob/b32ab427126220261bc3b4b261e0b16d15294d65/src/System.Web.Http.Cors/CorsMessageHandler.cs#L50-L64
It would be one thing if it were only catching exceptions for its own logic, but it's also catch exceptions from any upstream handlers.

Fixing it would have to be opt-in config to avoid breaking people.

dougbu:
I'll look at this and see what the approach might look like. We can decide based on that&hellip;

itrofimow:
You guys are the best.

itrofimow:
Hey guys. Any recent activity on this?
I might submit a PR if someone shares with me the design solution

mkArtakMSFT:
@dougbu, can you please share your findings here?

itrofimow:
Hi.
Any recent activity on this?
I mean its been a while, may be i can help somehow?

dougbu:
@itrofimow we have been focused on other priorities. What's needed here is an extensibility point to more easily override or hook into Cors exception handling. I haven't looked closely but other approaches might also solve the issue.

If you would like to give it a shot, feel free. Please open your PR early so that we can provide feedback on your design while it's a work in progress.

itrofimow:
Yeah, sure. Thanks

itrofimow:
@dougbu ping

[...]

PR #185, opened 2018-08-08
Fix for #147

itrofimow:
Adresses #147

### DIFF
--- a/src/System.Web.Http.Cors/CorsHttpConfigurationExtensions.cs
+++ b/src/System.Web.Http.Cors/CorsHttpConfigurationExtensions.cs
@@ -24,10 +24,29 @@ namespace System.Web.Http
         /// Enables the support for CORS.
         /// </summary>
         /// <param name="httpConfiguration">The <see cref="HttpConfiguration"/>.</param>
+        public static void EnableCors(this HttpConfiguration httpConfiguration)
+        {
+            EnableCors(httpConfiguration, null, false);
+        }
+        
+        /// <summary>
+        /// Enables the support for CORS.
+        /// </summary>
+        /// <param name="httpConfiguration">The <see cref="HttpConfiguration"/>.</param>
         /// <param name="rethrowExceptions">Indicates whether upstream exceptions should be rethrown</param>
-        public static void EnableCors(this HttpConfiguration httpConfiguration, bool rethrowExceptions = false)
+        public static void EnableCors(this HttpConfiguration httpConfiguration, bool rethrowExceptions)
         {
             EnableCors(httpConfiguration, null, rethrowExceptions);
+        }
+
+        /// <summary>
+        /// Enables the support for CORS.
+        /// </summary>
+        /// <param name="httpConfiguration">The <see cref="HttpConfiguration"/>.</param>
+        /// <param name="defaultPolicyProvider">The default <see cref="ICorsPolicyProvider"/>.</param>
+        public static void EnableCors(this HttpConfiguration httpConfiguration, ICorsPolicyProvider defaultPolicyProvider)
+        {
+            EnableCors(httpConfiguration, defaultPolicyProvider, false);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


dougbu:
@Tratcher I would appreciate you giving this a look. Seems reasonable to me though it adds a few more overloads to the public surface. Any concerns?

src/System.Web.Http.Cors/CorsMessageHandler.cs
```diff
@@ -60,6 +72,9 @@ protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 }
                 catch (Exception exception)
                 {
+                    if (_rethrowExceptions)
+                        throw;
```
Tratcher: style: include the braces please.

test/System.Web.Http.Cors.Test/CorsMessageHandlerTest.cs
```diff
@@ -7,13 +7,22 @@
...
 using System.Web.Cors;
+using System.Web.Http.ExceptionHandling;
 using System.Web.Http.Hosting;
 using Microsoft.TestCommon;
 
 namespace System.Web.Http.Cors
 {
     public class CorsMessageHandlerTest
     {
+        private class PassthroughExceptionHandler : IExceptionHandler
```
itrofimow: I feels like it's not the best place for this class, however i have no idea where it should be placed(
Any suggestions would be appreciated

test/System.Web.Http.Cors.Test/CorsMessageHandlerTest.cs
```diff
@@ -180,6 +189,40 @@ public async Task SendAsync_HandlesExceptions_ThrownDuringPreflight()
...
+
+            await Assert.ThrowsAsync<HttpResponseException>(() => invoker.SendAsync(request, CancellationToken.None));
+        }
+
+        [Fact]
+        public async Task SendAsync_RethrowsExceptions_WhenRethrowFlagIsTrue()
+        {
+            HttpConfiguration config = new HttpConfiguration();
+            config.Routes.MapHttpRoute("default", "{controller}");
+            config.Services.Replace(typeof(IExceptionHandler), new PassthroughExceptionHandler());
```
itrofimow: This is needed to reproduce the exact scenario of original issue, otherwise default implementation of IExceptionHandler maps Exceptions to just 500 status code

test/System.Web.Http.Cors.Test/CorsMessageHandlerTest.cs
dougbu: Bottom of this class is fine.

itrofimow:
I added tests and braces.

src/System.Web.Http.Cors/CorsHttpConfigurationExtensions.cs
```diff
@@ -26,16 +26,38 @@ public static class CorsHttpConfigurationExtensions
         /// <param name="httpConfiguration">The <see cref="HttpConfiguration"/>.</param>
         public static void EnableCors(this HttpConfiguration httpConfiguration)
         {
-            EnableCors(httpConfiguration, null);
+            EnableCors(httpConfiguration, null, false);
```
dougbu: FYI we've got a meeting scheduled next Tuesday to decide whether default should be `true` i.e. whether we should break compatibility to ensure `IExceptionHandler` sees what it should. Will let you know&hellip;

itrofimow: Thanks.

test/System.Web.Http.Cors.Test/CorsMessageHandlerTest.cs
itrofimow: Done

src/System.Web.Http.Cors/CorsHttpConfigurationExtensions.cs
dougbu: Decision was to avoid breaking changes and stick with the approach you've implemented here.

Issue #187, opened 2018-08-10
System.Net.Http.FormattingUtilities missed  RFC 850 


bhukivskyi:
System.Net.Http.FormattingUtilities.TryParseDate is used for cookie expire date parsing, but doesn't support RFC 850  (`ddd, d-MMM-yyyy H:m:s 'GMT`') format. Same issue in Core was fixed by https://github.com/aspnet/HttpAbstractions/pull/897/commits/b41d8656d5c6c522f9966793214b67753b24331a

dougbu:
@bhukivskyi thank-you for reporting this issue. We do not think this is a priority at the moment.

Feel free to submit a PR to fix this.

TKharaishvili:
PR - https://github.com/aspnet/AspNetWebStack/pull/196

dougbu:
e0f1f1338e

PR #196, opened 2018-09-18
FormattingUtilities support for parsing dates in RFC 850 long year format

TKharaishvili:
Addresses https://github.com/aspnet/AspNetWebStack/issues/187


TKharaishvili:
@dougbu added the change you requested. Thanks for the guidance.

PR #207, opened 2018-11-07
Supporting open generics in Razor Views

mansoor-omrani:
New requested feature: Supporting Generic Models in Razor Views

**Steps:**
- MVC source code changes: Implemented
- Razor Engine soure code changes: Not implemented (Too Hard)

I explained feature details in [https://github.com/aspnet/Mvc/issues/7152](https://github.com/aspnet/Mvc/issues/7152).

I believe the changes I applied in MVC soure code is sufficient, (it doesn't break existing MVC and at the same times provides this new feature). but in order for it to be working, other changes are needed to be applied in Razor Engine which were too hard for me to be done.

I was working on this feature last year until I reached to Razor and stuck. So, I gave up.

I thought to share my code changes, maybe somebody can complete the job.


mansoor-omrani:
I could track down codes in System.Web.Razor project to see where the VB/C# source code is generated for a view.

It seems that the classname of a razor view is set in 'WebPageRazorHost'. This class in turn is instantiated by 'WebRazorHostFactory'.

If we can could somehow pass generic types list to WebRazorHostFactory, I think the job will be complete.

The only remaining thing is to change 'SanitizeClassName()' method in ParserHelpers class (add a Type[] array parameter containing generic types) so that it returns class name of a view with generic arguments (like "_Admin_Area_Blogs_Index_csharp<T>").

I feel though that more changes are required since the classname will no longer be a solid name (it will contain generic characters <>).

Perhaps more changes will also be needed in GenerateCodeCore() method in RazorTemplateEngine class to handle this.

dougbu:
Thank-you for your proposed contribution. This feature is not something we plan to invest in for aspnet/AspNetWebStack.

Issue #203, opened 2018-10-19
Whitespace (any uri escapable char) in virtual path root results in NoRouteFound in an WebApi/Owin setup


adzhiljano:
Hello,

First of all, thank you very much for the Asp.Net and Katana!!!

I'm not entirely sure if this is an issue with WebApi or Katana or with the integration bits.

The environment is the following:

`Packages` (should be latest possible versions):

```
  <package id="Microsoft.AspNet.WebApi" version="5.2.6" targetFramework="net461" />
  <package id="Microsoft.AspNet.WebApi.Client" version="5.2.6" targetFramework="net461" />
  <package id="Microsoft.AspNet.WebApi.Core" version="5.2.6" targetFramework="net461" />
  <package id="Microsoft.AspNet.WebApi.Cors" version="5.2.6" targetFramework="net461" />
  <package id="Microsoft.AspNet.WebApi.Owin" version="5.2.6" targetFramework="net461" />
  <package id="Microsoft.AspNet.WebApi.WebHost" version="5.2.6" targetFramework="net461" />
  <package id="Microsoft.Owin" version="4.0.0" targetFramework="net461" />
  <package id="Microsoft.Owin.Cors" version="4.0.0" targetFramework="net461" />
  <package id="Microsoft.Owin.Host.SystemWeb" version="4.0.0" targetFramework="net461" />
  <package id="Microsoft.Owin.Security" version="3.1.0" targetFramework="net461" />
  <package id="Microsoft.Owin.Security.OAuth" version="3.1.0" targetFramework="net461" />
  <package id="Owin" version="1.0" targetFramework="net461" />
```
`Hosting`: IIS 10
`OS`: Win 10
`Site/App`: 'Virtual' application under the `Default Web Site (localhost:80)` with the name `test api`

But can be reproduced with the following minimal setup:
```
public void Configuration(IAppBuilder app)
{
	HttpConfiguration config = new HttpConfiguration();

	config.MapHttpAttributeRoutes();

	app.UseWebApi(config);
}
```

When I try to send a request to `http://localhost/test%20api/` I get a 404 and this is what I managed to trace:

1. The Uri related contents of the OwinRrequest (got it from a fake first middleware in the pipeline)
![1](https://user-images.githubusercontent.com/5879092/47230285-86070500-d3d2-11e8-996e-78291c01bb9b.JPG)
[...]

PR #211, opened 2018-11-19
Make `OwinHttpRequestContext.VirtualPathRoot` unescaped

dougbu:
- #203
- now consistent with other overrides of this property

### DIFF
--- a/test/System.Web.Http.Owin.Test/OwinHttpRequestContextTests.cs
+++ b/test/System.Web.Http.Owin.Test/OwinHttpRequestContextTests.cs
@@ -564,6 +564,30 @@ namespace System.Web.Http.Owin
             }
         }
 
+        // Regression test for #203.
+        [Fact]
+        public void VirtualPathRootGet_ReturnsUnescapedContextRequestPathBase()
+        {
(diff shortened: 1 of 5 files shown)


test/System.Web.Http.Owin.Test/OwinHttpRequestContextTests.cs
```diff
@@ -564,6 +564,30 @@ public void VirtualPathRootGet_ReturnsContextRequestPathBase()
...
+        // Regression test for #203.
+        [Fact]
+        public void VirtualPathRootGet_ReturnsUnescapedContextRequestPathBase()
+        {
+            // Arrange
+            var expectedVirtualPathRoot = "/a b";
+            var owinRequestMock = new Mock<IOwinRequest>(MockBehavior.Strict);
+            owinRequestMock
+                .Setup(r => r.PathBase)
+                .Returns(new PathString(expectedVirtualPathRoot));
```
dougbu: This test is a `[Fact]` because `PathString` throws if virtual path root is escaped or doesn't start with `'/'`.

Issue #206, opened 2018-11-06
`ByteRangeStream` does not limit `CopyToAsync(...)` range as it should


dougbu:
Fix is to _remove_ the `CopyToAsync(...)` override in `DelegatingStream`. The underlying `Stream`'s implementation of this method will do the **Right Thing:tm:** and use methods that are correctly overridden in `DelegatingStream` and its subclasses, including `ByteRangeStream`.

Issue originally tracked in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/690342

dougbu:
39d3064baf

PR #213, opened 2018-11-25
Flesh out `ByteRangeStream` methods to make it always limit reads

dougbu:
- #206
- `Seek(...)` was a no-op, `Position` was incorrect if `_lowerbounds != 0`, `ReadAsync(...)` read entire inner `Stream`
  - rewrite `ByteRangeStreamTest` to cover new `ByteRangeStream` members and hold fewer resources
- remove `DelegatingStream.CopyToAsync(...)` override because it copied entire inner `Stream` and was not needed
  - base implementation invokes `ReadAsync(...)`

### DIFF
--- a/src/System.Net.Http.Formatting/Internal/ByteRangeStream.cs
+++ b/src/System.Net.Http.Formatting/Internal/ByteRangeStream.cs
@@ -3,12 +3,14 @@
 
 using System.IO;
 using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
 using System.Web.Http;
 
 namespace System.Net.Http.Internal
 {
     /// <summary>
-    /// Stream which only exposes a read-only only range view of an 
+    /// Stream which only exposes a read-only only range view of an
     /// inner stream.
     /// </summary>
     internal class ByteRangeStream : DelegatingStream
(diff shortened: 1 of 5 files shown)


dougbu:
FYI @stephentoub

src/System.Net.Http.Formatting/Internal/ByteRangeStream.cs
```diff
@@ -132,33 +182,49 @@ public override void EndWrite(IAsyncResult asyncResult)
...
-        /// Gets the 
+        /// Gets the correct count for the next read operation.
         /// </summary>
         /// <param name="count">The count requested to be read by the caller.</param>
         /// <returns>The remaining bytes to read within the range defined for this stream.</returns>
         private int PrepareStreamForRangeRead(int count)
         {
-            long effectiveCount = Math.Min(count, _totalCount - _currentCount);
-            if (effectiveCount > 0)
+            // A negative count causes base.Raad* methods to throw an ArgumentOutOfRangeException.
```
dougbu: Rewrote method because `_currentCount` may now exceed `_totalCount` and we need to distinguish that case (a no-op) from `count < 0` (an error).

dougbu:
:up::date: to rebase and address @mkArtakMSFT's comment.
Will merge once the CI check passes.

PR #273, opened 2020-03-16
Update TempDataDictionary.cs

ThibaultLesuisse:
TryAndGetValue already remove the key from the _initialKeys HashSet,


dougbu:
@ThibaultLesuisse thank you for this contribution. We are currently only fixing critical functional issues in the code base, and have decided to not take this change at this time. (Removing a non-existent object from a `HashSet<T>` just causes the `Remove` method to return `false`.)

Issue #263, opened 2019-10-22
HttpResponseMessage Serialisation still breaks in .NET Core 3.0


aliostad:
This is related to https://github.com/dotnet/corefx/issues/31918 and the same repro described there breaks with .NET 3.0 but the reason appears to be different and due to not using `TryAddWithoutValidation` according to [this](https://github.com/dotnet/corefx/issues/31918#issuecomment-543414911).

Basic issue is that headers such as `Expires: -1` are not supported while they are still prevalent.

aliostad:
![tenor](https://user-images.githubusercontent.com/487334/67485045-37cdb700-f661-11e9-945d-7ec79a0c0bcf.gif)

jkonecki:
This issue has been present since .NET Core 2.1!

@aliostad keeps opening new bugs after we are told the issue has been fixed. Could someone PLEASE take a look at it and sort it out once and for all?!

dougbu:
c55dd95aebb2

PR #302, opened 2021-07-29
[releaes/3.2.8] Get builds working again

dougbu:
- test using netcoreapp2.1
- react to .NET Core breaking change
  - see https://docs.microsoft.com/en-us/dotnet/core/compatibility/2.1#path-apis-dont-throw-an-exception-for-invalid-characters
- suppress NU5125 warnings
  - break the build because warnings are treated as errors
- do not validate Expires header values
  - #263


Pilchie (approved):
Seems reasonable to me.

dougbu:
Hearing no objections&hellip;

Issue #307, opened 2021-10-05
Increase HTTP header line size limit during deserialization of HttpRequestMessage


aturabi:
### Getting "_Error parsing HTTP message header byte 2048 of message System.Byte[]_" error when deserializing HttpRequestMessage, asking to get HTTP header line size limit to match whatever limits are in place for HttpRequestMessage and serialization

When calling the method 'HttpContentMessageExtensions.ReadAsHttpRequestMessageAsync' (https://docs.microsoft.com/en-us/previous-versions/aspnet/hh834457(v%3dvs.118)) in System.Net.Http.Formatting to deserialize an HTTP message, I am sometimes getting the error "_Error parsing HTTP message header byte 2048 of message System.Byte[]_". 

This HTTP header line size is not a value that can be adjusted by any of the method's input parameters.

If this value is going to be hard-coded (and the user is not able to pass them in), it would be great if it could at least honor the HTTP header line size limits for the actual HTTP request message object, and the limits for the serialization. (Why are we able to populate an obj with HTTP headers this size, and serialize it without any error but deserialization gets an error?)

Someone else with same issue: https://forums.asp.net/t/2037342.aspx?Max+header+size+when+serializing+HttpMessageContent

PR #312, opened 2022-01-14
Configure max request line size limits to be the same as maxHeaderSize

pranavkm:
As part of ReadAsHttpRequestMessageAsync, the parsing of the HTTP request line is limited
by a non-configurable 2k limit. The size does not affect buffer sizes, only the maximum allowed length.

This PR updates the ReadAsHttpRequestMessageAsync API to use the same limits for HTTP request line
as the HTTP header line, the latter which is configurable by user code.
In the default case, this means the HTTP request line size now supports a 16k limit before it throws.

Fixes https://github.com/aspnet/AspNetWebStack/issues/307

### DIFF
--- a/src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
+++ b/src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
@@ -223,7 +223,7 @@ namespace System.Net.Http
 
             HttpUnsortedRequest httpRequest = new HttpUnsortedRequest();
             HttpRequestHeaderParser parser = new HttpRequestHeaderParser(httpRequest,
-                HttpRequestHeaderParser.DefaultMaxRequestLineSize, maxHeaderSize);
+                maxHeaderSize, maxHeaderSize);
             ParserState parseStatus;
 
             byte[] buffer = new byte[bufferSize];
--- a/src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
+++ b/src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
@@ -223,7 +223,7 @@ namespace System.Net.Http
 
             HttpUnsortedRequest httpRequest = new HttpUnsortedRequest();
             HttpRequestHeaderParser parser = new HttpRequestHeaderParser(httpRequest,
-                maxHeaderSize, maxHeaderSize);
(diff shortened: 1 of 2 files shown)


test/System.Net.Http.Formatting.Test/HttpContentMessageExtensionsTests.cs
```diff
@@ -482,16 +482,50 @@ public Task ReadAsHttpResponseMessageAsync_LargeHeaderSize()
...
+                string.Format("Cookie: {0}", cookieValue),
+            };
+
+            HttpContent content = CreateContent(true, request, "sample body");
+            var httpRequestMessage = await content.ReadAsHttpRequestMessageAsync(Uri.UriSchemeHttp, 64 * 1024, 64 * 1024);
+
+            Assert.Equal(HttpMethod.Get, httpRequestMessage.Method);
+            Assert.Equal("/", httpRequestMessage.RequestUri.PathAndQuery);
+            Assert.Equal("msdn.microsoft.com", httpRequestMessage.Headers.Host);
+            IEnumerable<string> actualCookieValue;
```
dougbu, line 500: bet you're missing `out var` 😀

dougbu (approved):

test/System.Net.Http.Formatting.Test/HttpContentMessageExtensionsTests.cs
pranavkm: I tried that and it does not work in C# 5. This is like getting -ve 💯 points.

src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
```diff
@@ -223,7 +223,7 @@ public static Task<HttpRequestMessage> ReadAsHttpRequestMessageAsync(this HttpCo
 
             HttpUnsortedRequest httpRequest = new HttpUnsortedRequest();
             HttpRequestHeaderParser parser = new HttpRequestHeaderParser(httpRequest,
-                HttpRequestHeaderParser.DefaultMaxRequestLineSize, maxHeaderSize);
+                maxHeaderSize, maxHeaderSize);
```
pranavkm, line 226:
```suggestion
                Math.Max(HttpRequestHeaderParser.DefaultMaxRequestLineSize, maxHeaderSize), maxHeaderSize);
```

TanayParikh (approved):

PR #325, opened 2022-05-13
Update HttpControllerHandler.cs

RyanLiu99:
httpResponseBase.StatusDescription  could have limitation of 512 chars. So limit it to avoid exception.


dougbu:
What problem are you trying to solve @RyanLiu99❔ It's unclear why _silently_ truncating `ReasonPhrase` when assigning to `StatusDescription` would be an improvement.

PR #356, opened 2022-08-01
Fix code bug

MrXhh:
Fix code bug

### DIFF
--- a/src/System.Web.Http/ModelBinding/ErrorParameterBinding.cs
+++ b/src/System.Web.Http/ModelBinding/ErrorParameterBinding.cs
@@ -20,7 +20,7 @@ namespace System.Web.Http.ModelBinding
         {
             if (message == null)
             {
-                throw Error.ArgumentNull(message);
+                throw Error.ArgumentNull("message");
             }
             _message = message;
         }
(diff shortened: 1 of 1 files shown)


dougbu:
@MrXhh could you please file an issue that we can reference in future release notes❔

@mkArtakMSFT we should take this one-line fix because it's a straight-up bug and won't cause breaking changes for our users.

dougbu (approved):
Thanks @MrXhh❕ We'll get this in once we have an issue to reference in the commit description.

src/System.Web.Http/ModelBinding/ErrorParameterBinding.cs
```diff
@@ -20,7 +20,7 @@ public ErrorParameterBinding(HttpParameterDescriptor descriptor, string message)
         {
             if (message == null)
             {
-                throw Error.ArgumentNull(message);
+                throw Error.ArgumentNull("message");
```
mkArtakMSFT, line 23: Should this be `nameof(message)` instead?
@dougbu  you're probably more familiar with the convention used in this repo.

dougbu: We don't use `nameof` at all in this repository. Most of the code here predates C# 6.

MrXhh:
Fix #357

PR #365, opened 2022-11-21
Enable CodeQl3000

wtgodbe:
Copied most of eng/common/templates from Arcade and added a basic build script. Test build here: https://dnceng.visualstudio.com/internal/_build/results?buildId=2050272&view=results


dougbu (approved):
This gets things going 😃

I suggest either subscribing to Arcade SDK updates (even if that's unused) and taking all of eng/common/ in this repo or doing something similar to https://github.com/dotnet/spa-templates/blob/main/.github/workflows/arcade-sync.yml (and related files) but synchronizing just eng/common/templates -- what you've taken so far.

PR #366, opened 2022-11-21
Enable public Azure DevOps builds

dougbu:
- add azure-pipelines.yml
  - build about the same matrix as we have on TeamCity in Main.Integration builds
- change build.cmd to enable overrides of its command line parameters

nits:
- touch up .codeql.yml to avoid test code and include portable projects

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,7 @@ jobs:
     ${{ else }}:
       name: NetCore1ESPool-Internal
       demands: ImageOverride -equals windows.vs2017.amd64
-  timeoutInMinutes: 90
+  timeoutInMinutes: 30
 
   strategy:
     matrix:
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,92 @@
...
+  value: true
+
+trigger: [main]
+pr: ['*']
+
+jobs:
+- job: build
+  displayName: Build
+  pool:
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
```
mkArtakMSFT, line 28: So is this what differentiates whether the build was kicked-off from AzDO vs TFS ?

mkArtakMSFT: Ok, based on the later comment I see that this is used to differentiate between internal and public builds

mkArtakMSFT (approved):

TanayParikh (approved):
Don't have too much context in this, but looks good overall

PR #368, opened 2022-11-23
Do not fail tests immediately

dougbu:
- combine `RegexReplace` into `PrintTestRunSummary` task
  - remove unused `WarnOnNoMatch` parameter; inline other parameters
  - capture test summary in `msbuild` logs
    - avoid `System.Console` use in the task
  - fail task if any tests fail
- collect `xunit` task exit codes in case of catastrophic problems (not test failures)
  - fail build if `PrintTestRunSummary` doesn't but errors occurred

also:
- do not test assemblies in parallel; reduce port contention issues
- avoid `CallTarget` task

nits:
- clean some trailing whitespace in C# sources
  - mostly files I expect to update soon
- reduce `NuGet.exe restore` verbosity; no longer debugging that
- add console summary when using build.cmd
- reduce duplication in `RestorePackages` target
  - only restore NetCore projects when `$(BuildPortable)` is `true`
  - restore src NetStandard project transitively through NetStandard.Test project
- pass more properties w/ `Restore` target invocation

### DIFF
--- a/Runtime.msbuild
+++ b/Runtime.msbuild
@@ -11,7 +11,7 @@
     <BuildPortable Condition=" '$(BuildPortable)' == '' ">true</BuildPortable>
     <BuildInParallel Condition=" '$(BuildInParallel)' == '' And $(MSBuildNodeCount) &gt; 1 ">true</BuildInParallel>
     <BuildInParallel Condition=" '$(BuildInParallel)' == '' ">false</BuildInParallel>
-    <TestInParallel Condition=" '$(TestInParallel)' == '' ">$(BuildInParallel)</TestInParallel>
+    <TestInParallel Condition=" '$(TestInParallel)' == '' ">false</TestInParallel>
     <TestResultsDirectory>$(MSBuildThisFileDirectory)bin\$(Configuration)\test\TestResults\</TestResultsDirectory>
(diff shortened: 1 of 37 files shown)


Runtime.msbuild
```diff
@@ -65,25 +65,24 @@
...
-      <_NuGetPackagesAndSolutions Include="src\System.Net.Http.Formatting.NetCore\packages.config;
-          test\System.Net.Http.Formatting.NetCore.Test\packages.config;
-          Runtime.sln" />
+      <_NuGetPackagesAndSolutions Include="Runtime.sln" />
 
       <!-- Avoid restoring RuntimePortable.sln directly. -->
       <_NuGetPackagesAndSolutions Include="src\System.Net.Http.Formatting.NetCore\packages.config;
           test\System.Net.Http.Formatting.NetCore.Test\packages.config"
         Condition=" '$(BuildPortable)' == 'true' " />
-      <_ProjectsToRestore Include="src\System.Net.Http.Formatting.NetStandard\System.Net.Http.Formatting.NetStandard.csproj;
```
dougbu, line 76: another nit: this project is restored transitively and it's more reliable not to kick that off at the same time as the test project

TanayParikh (approved):

PR #370, opened 2022-11-23
Use Newtonsoft.Json.Bson except in NetCore projects

dougbu:
- avoid obsolete `BsonReader` and `BsonWriter` classes
  - use `NETFX_CORE` for remaining legacy case (for now)
- remove `NEWTONSOFTJSON10` define; need only `NETFX_CORE` now

### DIFF
--- a/test/System.Net.Http.Formatting.Test/Formatting/BsonMediaTypeFormatterTests.cs
+++ b/test/System.Net.Http.Formatting.Test/Formatting/BsonMediaTypeFormatterTests.cs
@@ -262,10 +262,10 @@ namespace System.Net.Http.Formatting
             BsonMediaTypeFormatter formatter = new BsonMediaTypeFormatter();
             HttpContent content = new StringContent(String.Empty);
             MemoryStream stream = new MemoryStream();
-#if NEWTONSOFTJSON10 // Json.NET 10's Bson package calculates the path in some exceptions differently.
-            string expectedPath = "Value";
-#else
+#if NETFX_CORE // Separate Bson package (not yet used in NETCore project) calculates the path in exceptions differently
             string expectedPath = string.Empty;
+#else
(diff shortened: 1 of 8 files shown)


test/System.Net.Http.Formatting.Test/Formatting/BsonMediaTypeFormatterTests.cs
```diff
@@ -262,10 +262,10 @@ public async Task FormatterThrowsOnWriteWithInvalidContent()
             BsonMediaTypeFormatter formatter = new BsonMediaTypeFormatter();
             HttpContent content = new StringContent(String.Empty);
             MemoryStream stream = new MemoryStream();
-#if NEWTONSOFTJSON10 // Json.NET 10's Bson package calculates the path in some exceptions differently.
-            string expectedPath = "Value";
-#else
+#if NETFX_CORE // Separate Bson package (not yet used in NETCore project) calculates the path in exceptions differently
             string expectedPath = string.Empty;
+#else
+            string expectedPath = "Value";
```
TanayParikh, line 268: Just to confirm, this inversion in logic was intentional?

dougbu: Yes. We have 3 `Http.Formatting` projects in src: `NetCore` for `Profile259` (which is supposed to be similar to `netstandard1.3` but is actually a superset in some ways), `NetStandard` for `netstandard2.0`, and "regular" for .NET Framework. Previously only the `NetStandard` project defined `NEWTONSOFTJSON10` and used the Json.Bson package. Now all but the `NetCore` project uses Json.Bson. That project defines `NETFX_CORE`.

TanayParikh (approved):

JamesNK (approved):

PR #371, opened 2022-11-23
Use .NET SDK in two test projects

dougbu:
Use .NET SDK in NetCore.Test project
- for now, unable to target `netcoreapp2.1` in NetCore.Test project
  - Microsoft.Net.Http is not compatible
  - System.Net.Http types have different identities, causing compilation failures
- switch to use `dotnet vstest` for any .NET Core assemblies (just one currently)
  - can use `Xunit` target for all .NET Framework assemblies
- make XunitXml.TestLogger available for .NET Core App testing
  - use above test logger in repo-level testing
  - integrate w/ tests that use the `Xunit` target
- don't import from xunit.msbuild.runner when targeting .NET Core; not needed
- add `$(CopyLocalLockFileAssemblies)` to these (now both .NET SDK) test projects
  - xUnit assemblies were not copied (when they should have been) w/o this

Use .NET SDK in Microsoft.TestCommon
- allows reference in .NET SDK test projects w/o duplicating package references
- not sure how NetStandard.Test targeted `netcoreapp2.1` before


test/Directory.Build.props
```diff
@@ -4,6 +4,9 @@
   <PropertyGroup>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
-    <TargetFrameworkVersion Condition="'$(MSBuildProjectName)' != 'System.Net.Http.Formatting.NetStandard.Test'">v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion
```
wtgodbe, line 7: Maybe change this to only fire if `TargetFrameworks` isn't already set - that way any future projects that switch to use `netcoreapp2.1` won't have to be added to this list

wtgodbe (approved):

test/Directory.Build.props
dougbu: Good idea. I'll try that out while trying to get this building on the CI.

dougbu: Doesn't work because this is a .props file and nothing sets `$(TargetFrameworkVersion)` in SDK-style projects. Also, some projects _read_ the property, meaning we can't use a .targets file.

test/System.Net.Http.Formatting.NetStandard.Test/System.Net.Http.Formatting.NetStandard.Test.csproj
```diff
@@ -1,17 +1,23 @@
...
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.11.0" />
     <PackageReference Include="Moq" Version="4.7.142.0" />
     <PackageReference Include="xunit" Version="2.3.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
```
wtgodbe, line 18: Why the spaces?

wtgodbe (approved):

test/System.Net.Http.Formatting.NetStandard.Test/System.Net.Http.Formatting.NetStandard.Test.csproj
dougbu: No idea. I just copied what NuGet.org recommended. See https://www.nuget.org/packages/xunit.runner.visualstudio

PR #372, opened 2022-11-28
Add workflow to sync eng/common/templates from arcade

wtgodbe:
In the style of https://github.com/dotnet/spa-templates/pull/59 (with a touch of https://github.com/dotnet/spa-templates/pull/61). @dougbu is anything else needed on the github end to enable this workflow?

### DIFF
--- a/.github/workflows/CopyToTarget.cmd
+++ b/.github/workflows/CopyToTarget.cmd
@@ -1,0 +1,19 @@
+@ECHO OFF
+SETLOCAL
+
+IF NOT [%1] == [] (set from=%1)
+IF [%from%] == [] (
+  echo The 'from' command line parameter is not set, aborting.
+  exit /b 1
+)
+
+IF NOT [%2] == [] (set to=%2)
+IF [%to%] == [] (
+  echo The 'to' command line parameter is not set, aborting.
+  exit /b 1
+)
+
+echo Copying from '%from%' to '%to'.
+
+REM https://superuser.com/questions/280425/getting-robocopy-to-return-a-proper-exit-code
+(robocopy %from% %to% /MIR) ^& IF %ERRORLEVEL% LSS 8 SET ERRORLEVEL = 0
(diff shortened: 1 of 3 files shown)


dougbu:
> @dougbu is anything else needed on the github end to enable this workflow?

Align <https://github.com/aspnet/AspNetWebStack/settings/actions> w/ settings from <https://github.com/dotnet/aspnetcore/settings/actions>. Probably don't need the peter-evans mention in the list of allowed reusable workflows.

dougbu (approved):
Should be fine for now. Note however that this repo needs absolutely nothing from eng/common/ at the moment. That'll change as soon as we start extending internal builds to e.g. sign or create SBOMs.

wtgodbe:
> Should be fine for now. Note however that this repo needs absolutely nothing from eng/common/ at the moment. That'll change as soon as we start extending internal builds to e.g. sign or create SBOMs.

I only added the sync for `/eng/common/templates`

wtgodbe:
> Align https://github.com/aspnet/AspNetWebStack/settings/actions w/ settings from https://github.com/dotnet/aspnetcore/settings/actions. Probably don't need the peter-evans mention in the list of allowed reusable workflows.

Done

dougbu:
> I only added the sync for `/eng/common/templates`

I know. We don't use anything from there either. The previous mention of pool-providers.yml didn't do anything because your pipeline didn't need or use the variables set in that file.

wtgodbe:
> I know. We don't use anything from there either. The previous mention of pool-providers.yml didn't do anything because your pipeline didn't need or use the variables set in that file.

Wow, I must be losing my mind - I could have sworn I was using `eng\common\templates\jobs\codeql-build.yml`. I'm more inclined to just hold off on this until we actually need eng\common stuff in that case, and just add an arcade subscription at that point - thoughts?

dougbu:
> I'm more inclined to just hold off on this until we actually need eng\common stuff in that case, and just add an arcade subscription at that point - thoughts?

Hmm, initially either was fine w/ me. But I seem to remember some parts of the Arcade templates e.g. steps that publish to the BAR require more than just YAML. Let's see what happens&hellip;

PR #375, opened 2022-12-12
Build w/ VS 2019; restore w/ `msbuild`

dougbu:
- handle both SDK-style and legacy projects in one way
  - set `$(RestorePackagesConfig)` to cover both project types
- allow newer .NET SDKs to align w/ VS components
- make a few `<import/>`s conditional to avoid problems in `Restore` target evaluation
- restore global packages.config before building to make StyleCop available
- move second copies of packages.config packages into packages/ folder
- update FxCop TFM
  - .NET 4.0 not supported because current FxCopSdk is built against .NET 4.7.2
- suppress new warnings about old FxCop approach and not using `IFormatProvider`


dougbu:
This should unblock #371

dougbu:
Rebased on main to correct conflicts. Should be ready to go&hellip;

wtgodbe (approved):

PR #378, opened 2023-01-20
Address a few more nits

dougbu:
- avoid repeated NuGet.exe downloads
  - remove extra bits in modern file versions before parsing
- suppress warnings about targeting `netcoreapp2.1`

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,5 +9,8 @@
       analyzers, which run during build. Refer to https://aka.ms/fxcopanalyzers to migrate to FxCop analyzers."
     -->
     <SuppressLegacyCodeAnalysisDeprecatedWarning>true</SuppressLegacyCodeAnalysisDeprecatedWarning>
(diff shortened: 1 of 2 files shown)


Directory.Build.props
```diff
@@ -9,5 +9,8 @@
       analyzers, which run during build. Refer to https://aka.ms/fxcopanalyzers to migrate to FxCop analyzers."
     -->
     <SuppressLegacyCodeAnalysisDeprecatedWarning>true</SuppressLegacyCodeAnalysisDeprecatedWarning>
+    <!-- Workaround continued use of netcoreapp2.1; suppress the NETSDK1138 warning. -->
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
```
wtgodbe, line 13: Worth conditioning this on `TFM=netcoreapp2.1`, in case somebody checks in a `netcoreapp1.0` target or something?

wtgodbe (approved):

Directory.Build.props
dougbu: I think we'd notice :grin:

PR #380, opened 2023-01-22
Move NetCore project to use .NET SDK

dougbu:
Move NetCore project to use .NET SDK
- Formatting.NetCore now targets `netstandard1.3`
  - was `Profile259` aka `portable-net45+wp80+win8+wpa81`, approximately `netstandard1.0`
  - but Newtonsoft.Json.Bson supports `netstandard1.3` and up
- use Newtonsoft.Json.Bson in NetCore projects
- can now multi-target NetCore.Test as well
- nits:
  - align src/ Formatting projects as much as possible
  - align test/ Formatting projects as much as possible
  - clean up the Microsoft.TestCommon project slightly
- note: NetCore project names make no sense but I'm leaving them alone for this PR

React to using Newtonsoft.Json.Bson everywhere
- remove deprecated `BsonWriter` mentions entirely
- remove legacy BSON writer test workaround

Remove unnecessary `ConcurrentDictionary` class
- was not available in `Profile259`
- leave tests to confirm correct operation w/ "real" class

Remove `UriQueryUtility` because `WebUtility` is available everywhere
- leave test to confirm `WebUtility` continues to behave about the same
  - handle `WebUtility` being a `public` class
- change other tests to handle `WebUtility.UrlEncode(...)` encoding to uppercase
  - unlike `UriQueryUtility.UrlEncode(...)`

Add simple classes to NetCore project
- use `ICloneable` copy to get more classes compiling
  - helps NetCore in similar way to existing `MediaTypeHeaderValueExtensions`
- use `InvalidEnumArgumentException` copy to simplify code and testing
- both not available in `netstandard1.3` (or available packages) but easily duplicated

Remove unnecessary NetCore-specific GlobalSuppressions.cs file

Use `NameValueCollection` everywhere
- available (but not serializable) in `netstandard1.3`

Add `MaxDepth` support in NetCore assembly
- remove an unnecessary suppression about `MaxDepth` in `BaseJsonMediaTypeFormatter`

Add `UseDataContractJsonSerializer` support in NetCore assembly
- `XsdDataContractExporter` remains unavailable
  - update XML DCS tests to handle this (previously not in NetCore assembly)
- try various workarounds
  - `ReadOnlyStreamWithEncodingPreamble` causes tests to hit dotnet/runtime#80160
  - restriction to UTF8 when writing is a significant issue

Copy `TranscodingStream` and related files from dotnet/runtime
- my runtime clone was at 88868b7a781f4e5b9037b8721f30440207a7aa42

Adjust `TranscodingStream` and related code to compile and run here
- add conditions to avoid references to version-specific API
- remove references to internal dotnet/runtime code
  - rewrite some code more generically if performance impact likely minor
- use Microsoft.TestCommon instead of directly relying on Xunit in tests
- use C# 9.0

Use `TranscodingStream` in JSON and XML formatters
- remove `ReadOnlyStreamWithEncodingPreamble` and its tests
  - was insufficient in general and doesn't work correctly in `netstandard1.3`
  - see dotnet/runtime#80160
- now support a much broader set of encodings in JSON and XML formatters
  - not just UTF8, UTF16BE, and UTF16LE
  - JSON formatter already supported every encoding when `UseDataContractJsonSerializer` was `false`
    - no longer have this downside when property is `true`
    - however, quotas remain unsupported in `netstandard1.3` version of this formatter when property is `true`
  - XML formatter requirement for an XML declaration in non-UTF8 content was general and is no longer an issue

Add `MediaTypeMappings` support in NetCore assembly

Add `IRequiredMemberSelector` support in NetCore assembly

Remove unnecessary NETFX_CORE workarounds
- use `DefaultContractResolver`, `GuidAttribute`, `TypeDescriptor`, `XmlElement`, `XmlNode` classes
- use `IRequiredMemberSelector` interface
- use `RequestMessage.CreateResponse()` extension method
  - all available in `netstandard1.3`
- `JsonContractResolver` use is something of a late Newtonsoft.Json version reaction
  - was held back by missing `IRequiredMemberSelector`

Work around missing features in `netstandard1.3`
- get classes now included in NetCore assembly compiling
- `SerializationAttribute` is not available
  - therefore `Exception` is not serializable
  - therefore `DefaultContractResolver.IgnoreSerializableAttribute` is not available
- `Stream.(Begin|End)(Read|Write)` and `Stream.Close()` are not available
- nit: clean up `MultipartFormDataStreamProvider` whitespace

Ensure a couple more assemblies are available
- don't seem to be copied into test folder in CI builds


dougbu:
This PR is pretty big and complicated. I recommend going through the commits individually and concentrating on areas you're at least slightly familiar with.

@phenning, I included you because I'm removing some comments you added. Note we already have a `netstandard2.0` assembly w/ none of the restrictions of the former Profile259 assembly.

@stephentoub and @GrabYourPitchforks, you're likely most interested in the first two of the three commits mentioning `TranscodingStream`, especially the middle one.

dougbu:
To help track which commits have been thoroughly checked, please tick those you are signing off on:
- [ ] Move NetCore project to use .NET SDK
- [ ] React to using Newtonsoft.Json.Bson everywhere
- [ ] Remove unnecessary `ConcurrentDictionary` class
- [ ] Remove `UriQueryUtility` because `WebUtility` is available everywhere
- [ ] Add simple classes to NetCore project
- [ ] Remove unnecessary NetCore-specific GlobalSuppressions.cs file
- [ ] Use `NameValueCollection` everywhere
- [ ] Add `MaxDepth` support in NetCore assembly
- [ ] Add `UseDataContractJsonSerializer` support in NetCore assembly
- [ ] Copy `TranscodingStream` and related files from dotnet/runtime
- [ ] Adjust `TranscodingStream` and related code to compile and run here
- [ ] Use `TranscodingStream` in JSON and XML formatters
- [ ] Add `MediaTypeMappings` support in NetCore assembly
- [ ] Add `IRequiredMemberSelector` support in NetCore assembly
- [ ] Remove unnecessary NETFX_CORE workarounds
- [ ] Work around missing features in `netstandard1.3`

wtgodbe (approved):
Approving for the subset of commits I checked off, largely based off of the changes seeming reasonable/consistent, and the CI passing

src/System.Net.Http.Formatting/Internal/TranscodingStream.cs
```diff
@@ -25,10 +31,14 @@ internal sealed class TranscodingStream : Stream
         private const int MinWriteRentedArraySize = 4 * 1024;
         private const int MaxWriteRentedArraySize = 1024 * 1024;
 
+        internal static readonly byte[] EmptyByteBuffer = new byte[0];
+        internal static readonly char[] EmptyCharBuffer = new char[0];
```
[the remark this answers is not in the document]
dougbu: Right now, this class is compiled into `net45`, `netstandard1.3`, and `netstandard2.0` assemblies. While it's possible we'll expand the package to include more assemblies, that's not in the cards today. Even if it was, we're not breaking compatibility by bumping our .NET Framework target anytime soon. And `net45` doens't have `Array.Empty<T>`.

Does `Array.Empty<T>` perform so much better that I should make this conditional❔

That said, I think I'm using these `static`s in test code and a near-term change will bring our tests to `net462`. No need for these outside our src/ projects.

src/System.Net.Http.Formatting/Internal/TranscodingStream.cs
```diff
@@ -437,19 +452,22 @@ async ValueTask<int> ReadAsyncCore(Memory<byte> buffer, CancellationToken cancel
...
+            => throw Error.NotSupported(Properties.Resources.NotSupported_UnseekableStream);
 
         public override void SetLength(long value)
-            => ThrowHelper.ThrowNotSupportedException_UnseekableStream();
+            => throw Error.NotSupported(Properties.Resources.NotSupported_UnseekableStream);
 
+#if NET6_0_OR_GREATER
         [StackTraceHidden]
+#else
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
dougbu: I thought the affect would be a cheap way to do something similar to `[StackTraceHidden]` where that's not available. I suspect I'm missing something important (likely related to expanding the IL in the calling methods). Should I just leave the `#else` part out❔

src/System.Net.Http.Formatting/Internal/TranscodingStream.cs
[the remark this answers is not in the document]
stephentoub: > Does Array.Empty<T> perform so much better that I should make this conditional❔

No, not worth it.

src/System.Net.Http.Formatting/Internal/TranscodingStream.cs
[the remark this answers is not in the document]
stephentoub: I see. yes, please remove the added aggressive inlinings.

dougbu:
Rebased and addressed @stephentoub's comments.

@stephentoub any other thoughts❔ If none, please check the non-`NameValueCollection` commits off in the earlier checklist that aren't checked yet. They're all `TranscodingStream` related but I understand if parts aren't of interest to you. We could find someone to double-check the project files for example (if you prefer).

Other reviewers: Please look at the `NameValueCollection` commit.

stephentoub:
> please check the non-NameValueCollection commits off in the earlier checklist that aren't checked yet. They're all TranscodingStream related but I understand if parts aren't of interest to you. We could find someone to double-check the project files for example (if you prefer).

I don't have rights to click those checkmarks.  I only reviewed the "Adjust TranscodingStream and related code to compile and run here" commit and it looked fine other than my feedback you addressed.

dougbu:
> I don't have rights to click those checkmarks.

That's a bit odd but thanks for letting me know where you'd focused. Much appreciated.

dougbu:
To other reviewers: I also ticked the checkboxes for "Copy `TranscodingStream` &hellip;" because that was just a literal `cp` from my dotnet/runtime clone and "!fixup! Remove added `AggressiveInlining` requests" because that was in response to @stephentoub's comments. This leaves us w/ 3 unreviewed commits.

tools/WebStack.settings.targets
```diff
@@ -36,8 +36,8 @@
     <StyleCopTreatErrorsAsWarnings Condition=" '$(StyleCopTreatErrorsAsWarnings)' == '' ">false</StyleCopTreatErrorsAsWarnings>
     <StyleCopEnabled Condition=" '$(StyleCopEnabled)' == '' ">false</StyleCopEnabled>
 
-    <!-- Disable C# 6.0 features in Visual Studio. -->
-    <LangVersion Condition=" '$(MSBuildProjectExtension)' == '.csproj' ">5</LangVersion>
+    <!-- Use latest C# compiler supported in Visual Studio 2019. -->
+    <LangVersion Condition=" '$(MSBuildProjectExtension)' == '.csproj' ">9.0</LangVersion>
```
wtgodbe, line 40: This is a pretty big jump, is anyone relying on anything related to our LangVersion?

test/System.Net.Http.Formatting.Test/System.Net.Http.Formatting.Test.csproj
```diff
@@ -32,7 +32,7 @@
     <Reference Include="System.Memory, Version=4.0.1.2, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
       <HintPath>..\..\packages\System.Memory.4.5.5\lib\netstandard2.0\System.Memory.dll</HintPath>
       <SpecificVersion>False</SpecificVersion>
-      <Private>False</Private>
+      <Private>True</Private>
```
wtgodbe, line 35: How does `Private=true` make this _more_ available? (Not blocking on this, just curious)

wtgodbe (approved):
Approving again for the last few commits I just looked at

tools/WebStack.settings.targets
dougbu: We used to rely on the low version for continued compatibility w/ an outdated Visual Studio version. Now that we require a minimum VS 2019 version, this is just fine. It's also needed to avoid major additional changes in `TranscodingStream` and so on.

test/System.Net.Http.Formatting.Test/System.Net.Http.Formatting.Test.csproj
dougbu: It corresponds to `CopyLocal=true`, meaning the file will definitely appear in the test projects' bin/ folder.

src/System.Net.Http.Formatting/UriExtensions.cs
```diff
@@ -18,21 +18,12 @@ namespace System.Net.Http
     [EditorBrowsable(EditorBrowsableState.Never)]
     public static class UriExtensions
     {
-#if NETFX_CORE
-        /// <summary>
-        /// Parses the query portion of the specified <see cref="Uri"/>.
-        /// </summary>
-        /// <param name="address">The <see cref="Uri"/> instance from which to read.</param>
-        /// <returns>A <see cref="HttpValueCollection"/> containing the parsed result.</returns>
-        public static HttpValueCollection ParseQueryString(this Uri address)
```
Tratcher, line 27: Public API change?

Tratcher (approved):
These TFM changes are breaking. Will this be a new major version?

I didn't review the tests.

src/System.Net.Http.Formatting/Formatting/JsonMediaTypeFormatter.cs
```diff
@@ -233,10 +213,25 @@ public override object ReadFromStream(Type type, Stream readStream, Encoding eff
...
+
+#if NETFX_CORE
+                using (innerStream)
                 {
-                    return dataContractSerializer.ReadObject(reader);
+                    // Unfortunately, we're ignoring _readerQuotas.
+                    return dataContractSerializer.ReadObject(innerStream);
                 }
+#else
+                // XmlReader will always dispose of innerStream when we dispose of the reader.
```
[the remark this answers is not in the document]
dougbu: `JsonReaderWriterFactory.CreateJsonReader(...)` returns `XmlDictionaryReader`. (That's derived from `XmlReader` and it's the base class that implements `Dispose()`.) I'll update the comment to say `XmlDictionaryReader` and the `var reader` just below to use the type as well.

src/System.Net.Http.Formatting/UriExtensions.cs
dougbu: Technically yes though `HttpValueCollection` now consistently derives from `NameValueCollection` in all of the Formatting assemblies and the returned `object` is actually unchanged. The same is true for `FormDataCollection.ReadAsNameValueCollection()` and `HttpContentFormDataExtensions.ReadAsFormDataAsync(...)` (two overloads) due to the removed `NETFX_CORE`-specific `using` removal in each class. Further, the surface area of `HttpValueCollection` is a touch different in the `netstandard1.3` assembly because it has a base class instead of implementing an interface directly and `Add(...)` is now `virtual`.

> These TFM changes are breaking. Will this be a new major version?

Due to the large number of changes in dependencies for different TFMs in the Formatting package, the major increase in surface area of the `netstandard1.3` assembly in there, and the changes noted here, I lean toward bumping the Formatting package's version to 6.0.0 but leaving other packages at 3.3.0 and 5.3.0 because the main changes elsewhere were to the Owin and Newtonsoft package version bumps.

Thoughts❔

dougbu:
@Tratcher I addressed your comment on the "Use `TranscodingStream` &hellip;" commit in the latest "!fixup!" commit. Would you say you're signing off on those two commits❔

Still one open question on versioning and public API surface to you @Tratcher

src/System.Net.Http.Formatting/UriExtensions.cs
Tratcher: Sounds good.

Tratcher:
I've reviewed everything but the tests. :shipit:

PR #383, opened 2023-01-30
Bump test project dependencies

dougbu:
Bump test project dependencies
- Castle.Core, Moq, and xUnit versions were all out of date
- hold xunit.runner.visualstudio version back in .NET SDK test projects
  - can be moved to latest version after we stop testing on netcoreapp2.1
- add missing xunit.analyzers reference to System.Web.Razor.Test project
- move all test projects to .NET v4.6.2 (a supported framework)
  - remove `netstandard` reference in System.Net.Http.Formatting.Test; not needed w/ new TFM & updated references
- further separate build of Microsoft.TestCommon project when invoked from NetCore.Test project
  - special case `RestorePackages` for this case
  - add System.Net.Http references to avoid conflicting versions e.g. src/ and test/ TFMs differ

React to changed xUnit APIs
- adjust Microsoft.TestCommon code
- nit: use `Array.Empty<byte>()` in `TranscodingStreamTests`
  - `TranscodingStream` `internal`s can be `private` instead

Resolve xUnit issues new analyzers find
- address xUnit2000 warnings
  - pass expected values to `Assert.Equal(...)` as correct (left) argument
- make generic method types explicit to avoid
  `error CS0121: The call is ambiguous between the following methods or properties: ...`
- note: cannot remove unnecessary xUnit1013 suppression
  - related bug (xunit/xunit#1466) apparently not fixed in 1.0.0 analyzers package
  - was xunit/xunit.analyzers#82 fix (in 2017) insufficient?

React to new Moq changes
- avoid `ProtectedMock\`1.ThrowIfPublicMethod(...)` `throw`ing
  - use new overloads introduced in the Moq 4.11.0 release
- adjust to Moq hiding the `ObjectProxy` type better
- update `ControllerContext` mocking to avoid NREs
  - setting `HttpContext.User` on the `Object` left `get` value `null` (did nothing)
- nit: use `SetupGet(...)` for another property

### DIFF
--- a/test/Microsoft.AspNet.Facebook.Test/Microsoft.AspNet.Facebook.Test.csproj
+++ b/test/Microsoft.AspNet.Facebook.Test/Microsoft.AspNet.Facebook.Test.csproj
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\xunit.runner.visualstudio.2.3.1\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.3.1\build\net20\xunit.runner.visualstudio.props')" />
-  <Import Project="..\..\packages\xunit.core.2.3.0\build\xunit.core.props" Condition="Exists('..\..\packages\xunit.core.2.3.0\build\xunit.core.props')" />
+  <Import Project="..\..\packages\xunit.runner.visualstudio.2.4.5\build\net462\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.4.5\build\net462\xunit.runner.visualstudio.props')" />
+  <Import Project="..\..\packages\xunit.core.2.4.2\build\xunit.core.props" Condition="Exists('..\..\packages\xunit.core.2.4.2\build\xunit.core.props')" />
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory),Runtime.sln))\tools\WebStack.settings.targets" />
   <PropertyGroup>
     <ProjectGuid>{C3BEF382-C7C4-454D-B017-1EAC03E9A82C}</ProjectGuid>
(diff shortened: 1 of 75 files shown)


test/Microsoft.AspNet.Facebook.Test/Microsoft.AspNet.Facebook.Test.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\xunit.runner.visualstudio.2.3.1\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.3.1\build\net20\xunit.runner.visualstudio.props')" />
-  <Import Project="..\..\packages\xunit.core.2.3.0\build\xunit.core.props" Condition="Exists('..\..\packages\xunit.core.2.3.0\build\xunit.core.props')" />
+  <Import Project="..\..\packages\xunit.runner.visualstudio.2.4.5\build\net462\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.4.5\build\net462\xunit.runner.visualstudio.props')" />
```
wtgodbe, line 3: Would it be straightforward to parameterize versions in this repo? Would make future changes easier. But if it's pretty involved then it's probably not worth it.

test/System.Net.Http.Formatting.NetCore.Test/System.Net.Http.Formatting.NetCore.Test.csproj
```diff
@@ -9,15 +9,22 @@
     <DefineConstants>$(DefineConstants);NETFX_CORE</DefineConstants>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+    <!-- Version 2.4.5 does not support netstandard2.1. -->
+    <_VSRunnerVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">2.4.3</_VSRunnerVersion>
```
wtgodbe, line 14: Why not just use 2.4.3 everywhere?

wtgodbe (approved):
MSBuild changes look reasonable

test/Microsoft.AspNet.Facebook.Test/Microsoft.AspNet.Facebook.Test.csproj
dougbu: It would be fairly involved because packages.config files can't use `msbuild` variables. Therefore, not worthwhile.

In addition, it's worth doing these changes manually (well, w/ a careful regular expression or two 😉) because it gives us an opportunity to check the assembly paths (in case new assemblies are available).

test/System.Net.Http.Formatting.NetCore.Test/System.Net.Http.Formatting.NetCore.Test.csproj
dougbu: My aim is to use the latest versions everywhere I could. I also plan to expand the testing matrix to include `net6.0` and / or `net7.0`. Bottom line, `netcoreapp2.1` is a special case and I'd rather make that clear.

TanayParikh (approved):

PR #387, opened 2023-02-17
Add `Content` to all `HttpResponseMessage`s

### LINKED ISSUE #386: Do not expose `HttpResponseMessage.Content == null` to user code
In recent (5.0 and newer) .NET runtimes, `Content==null` is impossible for user code to see. Even in older runtimes, `HttpClient` would never leave `Content==null` either.
- see dotnet/runtime@b48900f3b37e6d68b5d84a3e48f0c5beb602a128 for the 5.0 change

Public code in this repo both exposes `Content==null` (inconsistently) and expects (mostly in tests) to see that state. Bad enough we expose that state but it being inconsistent (depending on the runtime version) makes things worse.

The action here is to change our code to both remove the inconsistencies i.e., have `Content!=null` everywhere regardless of the runtime version and update our expectations to handle `Content!=null`. This includes reverting tests removed on `net6.0` in #384.

dougbu:
- fix #386
  - reenable `net6.0` tests disabled in #384 for this issue
- also, check `Stream.CanSeek` less when calculating `HttpMessageContent.ContentLength` value
  - use inner `HttpContent.Headers.ContentLength` if available
  - this changes behaviour slightly for both requests and responses
    - observable as `HttpMessageContent.Headers.ContentLength!=null` for empty messages
  - for responses, avoids `ContentLength==null` versus `ContentLength==0` inconsistencies (depending on platform)
  - `ContentLength==null` may still occur in some corner cases (which existed before)
    - e.g. when inner `HttpContent` doesn't know its length and `ReadAsStreamAsync()` hasn't completed
- main change means `HttpResponseMessage`s we expose to user code are consistent across platforms
  - note: user code won't see `EmptyContent` in `HttpResponseMessage`s we create

### DIFF
--- a/src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
+++ b/src/System.Net.Http.Formatting/HttpContentMessageExtensions.cs
@@ -451,7 +451,7 @@ namespace System.Net.Http
             Contract.Assert(destination != null, "destination headers cannot be null");
             Contract.Assert(contentStream != null, "contentStream must be non null");
             HttpContentHeaders contentHeaders = null;
-            HttpContent content = null;
+            HttpContent content;
 
             // Set the header fields
             foreach (KeyValuePair<string, IEnumerable<string>> header in source)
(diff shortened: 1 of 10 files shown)


Tratcher (commented):
This version looks fine, but you say you're still seeing other issues?

Tratcher (approved):

PR #390, opened 2023-02-23
Correct `ValidateStreamForReading(...)`

### LINKED ISSUE #389: `MessageHttpContext` does not handle unreadable/unseekable `Stream`s well
As mentioned in hidden comments in #384 (because I changed the relevant lines slightly), `HttpMessageContent.ValidateStreamForReading(...)` uses `CanRead` where `CanSeek` would be more appropriate. This causes problems when the inner `HttpContent.ReadAsStreamAsync()` returns a `Stream` that is readable but not seekable. `EmptyContent` provides such a `Stream`. Specifically, error messages are unintelligible.

1. If the inner `HttpContent` provides an unreadable `Stream`, `HttpMessageContent.ValidateStreamForReading(...)` ignores the problem entirely. This allows execution to continue past where problems should have been detected.
  ``` text
  System.NotSupportedException: Stream does not support reading.
   at System.IO.StreamHelpers.ValidateCopyToArgs(Stream source, Stream destination, Int32 bufferSize)
   at System.IO.Stream.CopyToAsync(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at System.IO.Stream.CopyToAsync(Stream destination, CancellationToken cancellationToken)
   at System.Net.Http.StreamToStreamCopy.CopyAsync(Stream source, Stream destination, Int32 bufferSize, Boolean disposeSource, CancellationToken cancellationToken)
  --- End of stack trace from previous location where exception was thrown ---
   at System.Net.Http.HttpContent.CopyToAsyncCore(ValueTask copyTask)
   at System.Net.Http.HttpMessageContent.SerializeToStreamAsync(Stream stream, TransportContext context) in ..\src\System.Net.Http.Formatting\HttpMessageContent.cs:line 194
   at System.Net.Http.HttpContent.CopyToAsyncCore(ValueTask copyTask)
  ...
  ```
2. If the inner `HttpContent` provides a readable but unseekable `Stream`, `HttpMessageContent.ValidateStreamForReading(...)` again misses the issue and attempts to use `stream.Position` instead of providing a useful message.
  ``` text
    System.NotImplementedException: The method or operation is not implemented.
   at System.Net.Http.HttpMessageContentTests.ReadOnlyStream.set_Position(Int64 value) in ..\test\System.Net.Http.Formatting.Test\HttpMessageContentTests.cs:line 496
   at System.Net.Http.DelegatingStream.set_Position(Int64 value)
   at System.Net.Http.HttpMessageContent.ValidateStreamForReading(Stream stream) in ..\src\System.Net.Http.Formatting\HttpMessageContent.cs:line 369
   at System.Net.Http.HttpMessageContent.SerializeToStreamAsync(Stream stream, TransportContext context) in ..\src\System.Net.Http.Formatting\HttpMessageContent.cs:line 193
   at System.Net.Http.HttpContent.CopyToAsyncCore(ValueTask copyTask)
  ...
  ```

dougbu:
- fix #389
- check `Stream` before first read attempt
- check `CanRead` (not `CanSeek`) before second or later read attempts
- add lots of tests of related scenarios
- nit: correct a recent `HttpContentMessageExtensions` comment

### DIFF
--- a/src/System.Net.Http.Formatting/HttpMessageContent.cs
+++ b/src/System.Net.Http.Formatting/HttpMessageContent.cs
@@ -359,12 +359,21 @@ namespace System.Net.Http
 
         private void ValidateStreamForReading(Stream stream)
         {
+            // Stream is null case should be an extreme, incredibly unlikely corner case. Every HttpContent from
+            // the framework (see dotnet/runtime or .NET Framework reference source) provides a non-null Stream
+            // in the ReadAsStringAsync task's return value. Likely need a poorly-designed derived HttpContent
+            // to hit this. Mostly ignoring the fact this message doesn't make much sense for the case.
+            if (stream is null || !stream.CanRead)
+            {
+                throw Error.NotSupported(Properties.Resources.NotSupported_UnreadableStream);
+            }
+
             // If the content needs to be written to a target stream a 2nd time, then the stream must support
             // seeking (e.g. a FileStream), otherwise the stream can't be copied a second time to a target
             // stream (e.g. a NetworkStream).
             if (_contentConsumed)
             {
-                if (stream != null && stream.CanRead)
+                if (stream.CanSeek)
                 {
                     stream.Position = 0;
                 }
--- a/src/System.Net.Http.Formatting/HttpMessageContent.cs
+++ b/src/System.Net.Http.Formatting/HttpMessageContent.cs
@@ -361,7 +361,7 @@ namespace System.Net.Http
         {
             // Stream is null case should be an extreme, incredibly unlikely corner case. Every HttpContent from
             // the framework (see dotnet/runtime or .NET Framework reference source) provides a non-null Stream
-            // in the ReadAsStringAsync task's return value. Likely need a poorly-designed derived HttpContent
+            // in the ReadAsStreamAsync task's return value. Likely need a poorly-designed derived HttpContent
             // to hit this. Mostly ignoring the fact this message doesn't make much sense for the case.
             if (stream is null || !stream.CanRead)
             {
--- a/test/System.Net.Http.Formatting.Test/HttpMessageContentTests.cs
+++ b/test/System.Net.Http.Formatting.Test/HttpMessageContentTests.cs
@@ -46,20 +46,34 @@ namespace System.Net.Http
             return httpResponse;
         }
 
-        private static async Task<string> ReadContentAsync(HttpContent content)
+        private static async Task<string> ReadContentAsync(HttpContent content, bool unBuffered = false)
         {
-            await content.LoadIntoBufferAsync();
+            if (unBuffered)
+            {
+                var stream = new MemoryStream();
+                await content.CopyToAsync(stream);
+                stream.Position = 0L;
 
-            return await content.ReadAsStringAsync();
+                // StreamReader will dispose of the Stream.
+                using var reader = new StreamReader(stream);
+
+                return await reader.ReadToEndAsync();
+            }
+            else
+            {
+                await content.LoadIntoBufferAsync();
+
+                return await content.ReadAsStringAsync();
+            }
         }
 
-        private static async Task ValidateRequest(HttpContent content, bool containsEntity)
+        private static async Task ValidateRequest(HttpContent content, bool containsEntity, bool unBuffered = false)
         {
             Assert.Equal(ParserData.HttpRequestMediaType, content.Headers.ContentType);
             long? length = content.Headers.ContentLength;
             Assert.NotNull(length);
 
-            string message = await ReadContentAsync(content);
+            string message = await ReadContentAsync(content, unBuffered);
             if (containsEntity)
             {
                 Assert.Equal(ParserData.HttpRequestWithEntity.Length, length);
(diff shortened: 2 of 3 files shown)


src/System.Net.Http.Formatting/HttpMessageContent.cs
```diff
@@ -359,12 +359,21 @@ private byte[] SerializeHeader()
 
         private void ValidateStreamForReading(Stream stream)
         {
+            // Stream is null case should be an extreme, incredibly unlikely corner case. Every HttpContent from
+            // the framework (see dotnet/runtime or .NET Framework reference source) provides a non-null Stream
+            // in the ReadAsStringAsync task's return value. Likely need a poorly-designed derived HttpContent
```
Tratcher, line 364:
```suggestion
            // in the ReadAsStreamAsync task's return value. Likely need a poorly-designed derived HttpContent
```

test/System.Net.Http.Formatting.Test/HttpMessageContentTests.cs
```diff
@@ -46,20 +46,34 @@ private static HttpResponseMessage CreateResponse(bool containsEntity)
...
         {
-            await content.LoadIntoBufferAsync();
+            if (unBuffered)
+            {
+                var stream = new MemoryStream();
+                await content.CopyToAsync(stream);
+                stream.Position = 0L;
+
+                // StreamReader will dispose of the Stream.
+                using var reader = new StreamReader(stream);
```
Tratcher, line 58: Why CopyToAsync vs ReadAsStreamAsync?
```suggestion
                // StreamReader will dispose of the Stream.
                using var reader = new StreamReader(await content.ReadAsStreamAsync());
```

Tratcher (approved):

test/System.Net.Http.Formatting.Test/HttpMessageContentTests.cs
dougbu: Good point.

nit: Since we're no longer allocating the `Stream`, the comment doesn't make that much sense anymore.
```suggestion
                using var reader = new StreamReader(await content.ReadAsStreamAsync());
```

dougbu: About to revert changes here. `ReadAsStreamAsync()` often returns a `LimitMemoryStream` and that can only be read once, invalidating many of the new tests in this PR.

dougbu: Oh, `ReadAsStreamAsync()` also buffers the `HttpContent`. That invalidates a few other tests and completely undermines the intent of the `unBuffered` argument to this test method.

dougbu: Fixed again

Tratcher: ReadAsStreamAsync shouldn't buffer...?

dougbu: > ReadAsStreamAsync shouldn't buffer...?

Maybe it shouldn't but it does. See the default `HttpContent` implementations where `ReadAsStreamAsync()` ➡️ `CreateContentReadStreamAsync()` ➡️ `LoadIntoBufferAsync()` in <https://github.com/microsoft/referencesource/blob/master/System/net/System/Net/Http/HttpContent.cs> and <https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs>.

Tratcher: Oh, you're supposed to override CreateContentReadStream/Async if you can produce a stream without buffering.

dougbu: This behaviour is what I saw in the tests; it's not about `HttpMessageContent`. Those tests mostly use `StreamContent` and I _think_ `StreamContent` itself overrides enough to avoid buffering. But one of the tests failed, clearly showing buffering. I don't remember which test unfortunately. (It got lost among the many failures due to the read-once behaviour.)

dougbu: I'm probably wrong here. `HttpMessageContent` overrides `SerializeToStreamAsync(Stream stream, TransportContext context)` but not `CreateContentReadStreamAsync()`. That may be why `ReadAsStreamAsync()` behaves badly. I can test an override but am unsure how signfiicant a bug this may be. Thoughts❔

Tratcher: It's just in-efficient. I don't know that people will be calling ReadAsStreamAsync much. The WebApi infrastructure uses SerializeToStreamAsync.

dougbu: > It's just in-efficient. I don't know that people will be calling ReadAsStreamAsync much. The WebApi infrastructure uses SerializeToStreamAsync(?).

That's _somewhat_ true, I see 8 `content.CopyToAsync()` and 5 `content.ReadAsStreamAsync()` calls in the src/ of this repo. `SerializeToStreamAsync()` is the main `virtual` (actually `protected abstract`) method that `CopyToAsync()` uses; `ReadAsStreamAsync()` also uses `CreateContentReadStreamAsync()`. If you reconsider, feel free to file a bug in this repo.

My current perspective is that overriding `CreateContentReadStreamAsync()` would be difficult to achieve the desired "better way" documented e.g. at <https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L544>
and <https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L555-L557>
because there's no one `Stream` or `byte` array to wrap. Would also need to consider the other `HttpContent` derived classes in this repo, almost all of which do _not_ override `CreateContentReadStreamAsync()` as well as the fact this would be a breaking change.

PR #395, opened 2023-03-04
Use `msbuild` from VS 2022 if available

dougbu:
- should ease local and TeamCity builds
- nit: switch to 64bit `msbuild`

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,5 +1,4 @@
 @echo off
-pushd %~dp0
 setlocal
 
 if exist bin goto Build
@@ -29,7 +28,7 @@ for /f "usebackq tokens=*" %%i in (`%vswhere% -version 16 -latest -prerelease -p
     -requires Microsoft.Net.Component.4.5.2.TargetingPack ^
     -requires Microsoft.Net.Component.4.6.2.TargetingPack ^
     -property installationPath`) do (
-  set InstallDir="%%i"
+  set "InstallDir=%%i"
 )
 
 if not DEFINED InstallDir (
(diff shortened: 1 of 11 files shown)


wtgodbe (approved):

dougbu:
@wtgodbe I normally don't worry too much about this repo's build and test flakiness. In this case however, I seem to have changed timings enough for some incorrect assemblies to land in `bin/$(Configuration)/Test/` (causing assembly load issues in Moq) and for significantly increased port reservation issues e.g. `System.ServiceModel.AddressAlreadyInUseException : HTTP could not register URL http://localhost:50235/ because TCP port 50235 is being used by another application.`

My efforts so far haven't worked and I may back out my latest (xUnit collection) commit. In the meantime, I'm going to try another thing or two. First, calling in the big guns&hellip;

----

@halter73 @Tratcher could you please have a look at the logic in the `PortReserver` class❔ I don't see anything obviously wrong in it and was surprised the xUnit collection approach didn't help much. I suspect there's something fundamentally wrong. Note however that the class iseems to be used correctly i.e., instances are disposed everywhere they should be.

dougbu:
I forgot to mention: The issues related to `PortReserver` aren't new in this PR. It's more that retrying builds doesn't seem to work these days.

dougbu:
Somewhat predictably, [build #20230306.4](https://dev.azure.com/dnceng-public/public/_build/results?buildId=195552&view=results) immediately after this PR validated on its first attempt for this iteration failed w/ lots of port reservation problems. Hopefully there's something in `PortReserver` that jumps out at you @Tratcher and @halter73.

PR #400, opened 2023-03-15
Separate Microsoft.TestCommon more

dougbu:
- pre-restore and prebuild the project before anything that depends on it
  - avoid references from other test projects doing any real work
- make ordering even more explicit
  - avoid confusion between batching builds and `$(BuildInParallel)`
- turns out this also reduces the number of double writes binary logs show significantly

### DIFF
--- a/Runtime.msbuild
+++ b/Runtime.msbuild
@@ -52,14 +52,26 @@
     <Exec Command='"$(NuGetExe)" restore .nuget\packages.config -PackagesDirectory packages -NonInteractive -Verbosity quiet' />
   </Target>
 
+  <ItemGroup>
+    <_Testing_NetStandard1_3 Include="true;false" />
+  </ItemGroup>
+
   <Target Name="RestorePackages" DependsOnTargets="DownloadNuGet">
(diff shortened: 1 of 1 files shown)


dougbu:
Unfortunately, the `PortReserver` issues are still hurting in AzDO builds. I don't see them nearly as much in TeamCity (oddly enough). /cc @halter73 @Tratcher

dougbu:
TeamCity builds were the ones where I mainly saw problems w/ Microsoft.TestCommon. Using this branch, TeamCity builds #561 and #563 of Main.Integration was successful.

Tratcher:
Trying to find a free port like that will always be racy. It's better for the server to retry until it finds a port.

dougbu:
> Trying to find a free port like that will always be racy. It's better for the server to retry binding until it finds a port.

Thanks @Tratcher. Doesn't help my frustration w/ getting this PR in :wink: May either force the merge or just keep retrying until things work. I'm also open to adding a tactical fix here. I'm fine no matter the choice but am not going to dive into finding that workaround myself.

PR #434, opened 2025-09-10
Converted the license file to markdown

jamshedd:


danmoseley:
It doesn't render quite right if you view it. Some areas are quoted/indented

jamshedd:
> It doesn't render quite right if you view it. Some areas are quoted/indented

I reviewed and it looks right to me :( Some areas were meant to be indented. Is the indentation that looks wrong different from the indentation in the attached HTML version?

[aspnetcomponent_rtw_ENU.htm](https://github.com/user-attachments/files/22258502/aspnetcomponent_rtw_ENU.htm)

danmoseley:
fair enough.

danmoseley (approved):

PR #437, opened 2025-09-18
Don't run tests in internal builds

wtgodbe:


danmoseley (approved):

danmoseley:
I guess this is OK to unblock scanners given we have no plans to change any of the code in this repo.

PR #447, opened 2026-05-27
Remove specific image demand

wtgodbe:


olfek:
@wtgodbe ---
Isn't this removing the whole version override rather than the demand for a specific version?
Mind you, I know know nothing about anything.

wtgodbe:
@olfek it's just saying that any image of `windows.vs2019.amd64` will do, rather than that specific image version.

